PR #6, opened 2022-08-01
Move over Ownerhandling from thetechnick/package-operator

garrettthomaskth:


internal/ownerhandling/annotation_test.go
```diff
@@ -0,0 +1,106 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"testing"
+)
+
+func TestIndexOf(t *testing.T) {
```
thetechnick, line 15: A common pattern for naming test functions would be Test<struct>_<method><specifier>.
So this test case could be named "TestOwnerStrategyAnnotation_indexOf" with special cases either handled by `t.Run` inside that test case or by creating more specific test cases like `TestOwnerStrategyAnnotation_indexOfNil`.

Another common pattern is to use table driven tests/test fixtures:
```go
tests := []struct {
	name	string
}{
	{
		name:	"case 1 description",
	},
	{
		name:	"case 2 description",
	},
}

for _, test := range tests {
	t.Run(test.name, func(t *testing.T) {}
}
```
Don't use a map here, because map access order is randomized, making test execution potentially flaky.

garrettthomaskth: Thank you! I used it in one of the tests :)

thetechnick (approved):

PR #12, opened 2022-08-11
Move over probe package

garrettthomaskth:


docs/api-reference.md
```diff
@@ -434,6 +446,7 @@ Used in:
 ### PackageProbeKindSpec
 
 Kind package probe parameters.
+selects objects based on Kinda and API Group.
```
garrettthomaskth, line 449:
```suggestion
selects objects based on Kind and API Group.
```

internal/probing/probe.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type Prober interface {
+	Probe(obj *unstructured.Unstructured) (success bool, message string)
+}
+
+type List []Prober
```
garrettthomaskth, line 15: This name might be a little bit confusing because of the `list.List` type. Maybe a different name would be better

thetechnick: When used it's always `probing.List`, also when using the `parse` methods, nobody will directly interact with this type anyway.
Let me make them package-private... :)

internal/probing/parse.go
```diff
@@ -0,0 +1,80 @@
...
+	"context"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	corev1alpha1 "package-operator.run/apis/core/v1alpha1"
+)
+
+// Parse takes a list of ObjectSetProbes commonly defined within a ObjectSet phase and
+// compiles a single Probe to test objects with.
```
garrettthomaskth, line 13:
```suggestion
// Parse takes a list of ObjectSetProbes (commonly defined within a ObjectSetPhaseSpec) and
// compiles a single Prober to test objects with.
```

internal/probing/parse.go
```diff
@@ -0,0 +1,80 @@
...
+		probe, err = ParseSelector(ctx, pkgProbe.Selector, probe)
+		if err != nil {
+			return nil, fmt.Errorf("parsing selector of probe #%d: %w", i, err)
+		}
+		probeList[i] = probe
+	}
+	return probeList, nil
+}
+
+// ParseSelector reads a corev1alpha1.ProbeSelector and wraps a Prober, only executing it only when the selector matches.
```
garrettthomaskth, line 28: I am not sure what `only executing it only when the selector matches.` means here. I think it can be removed

internal/probing/probe.go
```diff
@@ -0,0 +1,133 @@
...
+		return false, fmt.Sprintf("%q missing", fe.FieldB)
+	}
+
+	if !equality.Semantic.DeepEqual(fieldAVal, fieldBVal) {
+		return false, fmt.Sprintf("%q != %q", fieldAVal, fieldBVal)
+	}
+	return true, ""
+}
+
+// StatusObservedGeneration wraps the given Prober and ensures that .status.observedGeneration is qual to .metadata.generation,
```
garrettthomaskth, line 117:
```suggestion
// StatusObservedGeneration wraps the given Prober and ensures that .status.observedGeneration is equal to .metadata.generation,
```

internal/probing/probe.go
```diff
@@ -0,0 +1,133 @@
...
+// before running the given probe. If the probed object does not contain the .status.observedGeneration field,
+// the given prober is executed directly.
+type StatusObservedGeneration struct {
```
garrettthomaskth, line 120: I might call this `ProberObservedGeneration` or `ObservedGenerationProber`, calling it a status makes it feel like a totally different kind of object, but it is actually just a specific type of prober

internal/probing/parse_test.go
```diff
@@ -0,0 +1,94 @@
...
+				},
+			},
+		},
+	}
+
+	p, err := Parse(ctx, osp)
+	require.NoError(t, err)
+	require.IsType(t, List{}, p)
+
+	if assert.Len(t, p, 1) {
```
garrettthomaskth, line 30: Wouldn't we ideally have a more specific test? For example:

```
	var probe Prober
	expected := List{&KindSelector{
		Prober: &StatusObservedGeneration{Prober: probe},
		GroupKind: schema.GroupKind{
			Group: osp[0].Selector.Kind.Group,
			Kind:  osp[0].Selector.Kind.Kind,
		},
	}}
	assert.True(t, reflect.DeepEqual(expected, p))
```

ps the example doesn't work and I don't know why :sweat_smile:

thetechnick: Comparing interface slices with different pointers is not super trivial.
`reflect.DeepEqual` is respecting types, so a `*StatusObservedGeneration` != `Prober` even though one can be assigned to the other.

`assert.Equal` should handle this nicer (and gives you output for debugging!), but in my tests it tested for pointer equality within the list, raising a test error and printing:
```
        	            	expected: probing.list{(*probing.kindSelector)(0xc0001b2cc0)}
        	            	actual  : probing.list{(*probing.kindSelector)(0xc0001b2c60)}
```
Which is correct.
It really depends on the context, whether lists containing pointers to different objects are considered equal.

If in doubt, test and type-cast explicitly.

garrettthomaskth: Ohh wow, go is a very unforgiving language compared to python :D Thank you for the thorough explanation!!

PR #24, opened 2022-08-29
Implement objectset controller

thetechnick:


thetechnick:
Ready for review, but not for merging.
Merging is blocked by resolving a few TODOs in devkube:
- [ ] adding support for kind export logs
- [ ] use logr helpers to put and get logger from contexts
- [ ] add "WaitToBeGone" or similar function to Waiter implementation.

internal/controllers/objectsets/revision_reconciler.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	if len(objectSet.GetPrevious()) == 0 {
+		// no previous revision(s) specified, default to revision 1
+		objectSet.SetStatusRevision(1)
+		return
+	}
+
+	// Determine new revision number by inspecting previous revisions:
+	revisions := make([]genericObjectSet, len(objectSet.GetPrevious()))
```
garrettthomaskth, line 39: Isn't it a little more efficient and less complex to do:
```
latestPreviousRevision := 0
....
if prevObjectSet.GetStatusRevision() > latestPreviousRevision:
    latestPreviousRevision = prevObjectSet.GetStatusRevision()
```

internal/controllers/objectsets/objectset_controller.go
```diff
@@ -0,0 +1,236 @@
...
+	}
+	return res, nil
+}
+
+// ensures the cache finalizer is set on the given object.
+func (c *GenericObjectSetController) ensureCacheFinalizer(
+	ctx context.Context, objectSet genericObjectSet,
+) error {
+	return controllers.EnsureCommonFinalizer(ctx, objectSet.ClientObject(), c.client, watchFinalizer)
+}
```
garrettthomaskth, line 192: It seems like sometimes it is called `cacheFinalizer` and sometimes it is called `watchFinalizer`. Are these two different things? Otherwise I think one name might be more clear

internal/controllers/controllers.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+type dynamicWatchFreer interface {
+	Free(ctx context.Context, obj client.Object) error
+}
+
+func HandleCommonDeletion(
+	ctx context.Context, obj client.Object,
+	c client.Client, dw dynamicWatchFreer,
+	cacheFinalizer string,
```
garrettthomaskth, line 41: Does `cacheFinalizer` really need to be a parameter? Isn't it static?

magefile.go
```diff
@@ -88,7 +88,7 @@ func (Test) Unit() error {
 }
 
 func (Test) Integration() error {
```
garrettthomaskth, line 90: I was not able to run this out of the box
```
Running target: Test:Integration
exec: go "test" "-v" "-failfast" "-count=1" "-timeout=20m" "./integration/..."
panic: creating runtime client: Get "<hive url>": dial tcp <ip-address>:443: i/o timeout

goroutine 1 [running]:
package-operator.run/package-operator/integration.init.0()
        /home/gathomas/GolandProjects/package-operator/package-operator/integration/environment.go:44 +0x168
FAIL    package-operator.run/package-operator/integration       30.067s
FAIL

```

internal/controllers/controllers.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+const (
+	// This label is set on all dynamic objects, to limit caches.
+	DynamicCacheLabel = "package-operator.run/cache"
+	// Revision annotations holds a revision generation number to order ObjectSets.
+	RevisionAnnotation = "package-operator.run/revision"
+)
+
+func EnsureCommonFinalizer(
```
garrettthomaskth, line 21: Is this supposed to be a generic function? If so, I think the paramter `cacheFinalizer` should be changed to `finalizer`

internal/controllers/objectsetphases/phase_reconciler.go
```diff
@@ -0,0 +1,482 @@
...
+	if err := r.dynamicCache.Watch(
+		ctx, owner.ClientObject(), desiredObj); err != nil {
+		return nil, fmt.Errorf("watching new resource: %w", err)
+	}
+
+	return r.reconcileObject(ctx, owner, desiredObj)
+}
+
+// Reads and builds an object as specified in a phase.
+// Including system labels, defaults and owner reference.
```
garrettthomaskth, line 191:
```suggestion
// Builds an object as specified in a phase.
// Includes system labels, namespace and owner reference.
```

internal/controllers/objectsetphases/phase_reconciler.go
```diff
@@ -0,0 +1,482 @@
...
+	if err != nil && errors.IsNotFound(err) {
+		// No matter who the owner of this object is,
+		// it's already gone.
+		return true, nil
+	}
+	if err != nil {
+		return false, fmt.Errorf("getting object for teardown: %w", err)
+	}
+
+	if !r.ownerStrategy.IsController(owner, currentObj) {
```
garrettthomaskth, line 149: Calling `desiredObject` in line 132 sets `owner` as the controller of `desiredObj`. Is it possible the `Get` call in line 138 overwrites this controllership?

internal/controllers/objectsetphases/phase_reconciler_test.go
```diff
@@ -0,0 +1,733 @@
...
+				{
+					Object: runtime.RawExtension{},
+				},
+			},
+		})
+		require.NoError(t, err)
+		assert.True(t, done)
+	})
+
+	t.Run("delete confirms", func(t *testing.T) {
```
garrettthomaskth, line 53: Could we think of a different name instead of "delete confirms". This is testing the case where the call to delete shows us that the object is already gone, right?

internal/controllers/objectsetphases/phase_reconciler_test.go
```diff
@@ -0,0 +1,733 @@
...
+			Objects: []corev1alpha1.ObjectSetObject{
+				{
+					Object: runtime.RawExtension{},
+				},
+			},
+		})
+		require.NoError(t, err)
+		assert.True(t, done)
+
+		// It's super important that we don't check ownership on desiredObj on accident, because that will always return true.
```
garrettthomaskth, line 96:
```suggestion
		// It's super important that we check IsController, not IsOwner, because IsOwner will always return true.
```

magefile.go
thetechnick: `./mage test:integration` needs a cluster already having PKO installed.
Try `./mage dev:integration` this will setup a new cluster with kind, compile and install PKO and then run the integration tests against it.

internal/controllers/objectsets/revision_reconciler.go
thetechnick: Yes, you are right. Change incoming :)

internal/controllers/objectsetphases/phase_reconciler_test.go
thetechnick: yes exactly, changing to: `already gone when deleting`?

internal/controllers/objectsetphases/phase_reconciler_test.go
thetechnick: I am more interested in the parameters given to `IsController`. It has to be called with `currentObj` and not `desiredObj`.
Trying to make the comment clearer... :)

internal/controllers/objectsetphases/phase_reconciler.go
thetechnick: Nope, the Get in line 138 is just using the key (Name/Namespace) and loads the result into `currentObj`.
`desiredObj` is not changed just read from.

internal/controllers/objectsets/objectset_controller.go
thetechnick: It was just the watch in the previous codebase and now we have a full cache implementation.
Will make sure cache is used everywhere - too much copy pasta :)

internal/controllers/objectsetphases/phase_reconciler.go
garrettthomaskth: Sorry, I don't think I was clear enough. `currentObj` is a deep copy of `desiredObj`, and `owner` is set to be the controller of `desiredObj` in the function `desiredObject` which is called in line 132. I am asking if it is possible for the Get in line 138 to overwrite the controller of `currentObj`, otherwise the condition of this if statement `!r.ownerStrategy.IsController(owner, currentObj` could never be true

internal/controllers/objectsetphases/phase_reconciler_test.go
garrettthomaskth: Yeah, maybe `already gone on delete`?

internal/controllers/objectsetphases/phase_reconciler.go
garrettthomaskth: Ohhhhhhhhhh I think this is what you were addressing in the comment in the test file "// It's super important that we don't check ownership on desiredObj on accident, because that will always return true."

internal/controllers/objectsetphases/phase_reconciler_test.go
garrettthomaskth: Ohhhhhhhhhhhhh now I get it :D

garrettthomaskth: I think I was confused because because we are checking if it is a controller, not directly checking ownership, right?

internal/controllers/objectsetphases/phase_reconciler_test.go
```diff
@@ -0,0 +1,733 @@
...
+			},
+		})
+		require.NoError(t, err)
+		assert.False(t, done) // wait for delete confirm
+
+		// It's super important that we don't check ownership on desiredObj on accident, because that will always return true.
+		ownerStrategy.AssertCalled(t, "IsController", owner, currentObj)
+	})
+
+	t.Run("not owner", func(t *testing.T) {
```
garrettthomaskth, line 150:
```suggestion
	t.Run("not controller", func(t *testing.T) {
```

internal/controllers/objectsetphases/phase_reconciler.go
```diff
@@ -0,0 +1,482 @@
...
+
+	// Take over object ownership by patching metadata.
+	if needsAdoption {
+		log := logr.FromContextOrDiscard(ctx)
+		log.Info("adopting object",
+			"OwnerKey", client.ObjectKeyFromObject(owner.ClientObject()),
+			"OwnerGVK", owner.ClientObject().GetObjectKind().GroupVersionKind(),
+			"ObjectKey", client.ObjectKeyFromObject(desiredObj),
+			"ObjectGVK", desiredObj.GetObjectKind().GroupVersionKind())
+		controllers.SetObjectRevision(updatedObj, owner.GetStatusRevision())
```
garrettthomaskth, line 288: Is this correct? If the revision number of the owner is less than the revision number of the object, isn't it possible to break the monotonically increase nature of revisions?

internal/controllers/objectsets/objectset_controller.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if !objectSet.ClientObject().GetDeletionTimestamp().IsZero() ||
+		objectSet.IsArchived() {
+		if err := c.handleDeletionAndArchival(ctx, objectSet); err != nil {
+			return ctrl.Result{}, err
+		}
+
+		objectSet.UpdateStatusPhase()
+		// this controller owns status alone, so we can always update it without optimistic locking.
+		objectSet.ClientObject().SetResourceVersion("")
```
garrettthomaskth, line 151: Why do we set the ResourceVersion to ""?

internal/ownerhandling/native.go
```diff
@@ -1,27 +1,68 @@
...
 	}
 	return false
 }
 
+func (s *OwnerStrategyNative) RemoveOwner(owner, obj metav1.Object) {
+	ownerRefs := obj.GetOwnerReferences()
+	foundIndex := -1
+	for i, ownerRef := range ownerRefs {
+		if owner.GetUID() == ownerRef.UID {
+			// remove owner
```
garrettthomaskth, line 56:
```suggestion
```

internal/ownerhandling/annotation.go
```diff
@@ -81,15 +90,43 @@ func (s *OwnerStrategyAnnotation) SetControllerReference(owner, obj metav1.Objec
...
 		}
 	}
 	return false
 }
 
+func (s *OwnerStrategyAnnotation) RemoveOwner(owner, obj metav1.Object) {
+	ownerRefs := s.getOwnerReferences(obj)
+	foundIndex := -1
+	for i, ownerRef := range ownerRefs {
+		if owner.GetUID() == ownerRef.UID {
```
garrettthomaskth, line 119: Some places the comparison is made with `referSameObject` and some places by comparing the UID. Is there are reason to have two different approaches?

internal/controllers/controllers.go
thetechnick: It can be static, but I found it best practice to name it by controller waiting on the finalizer:
e.g. `objectset.package-operator.run/cache`, `objectdeployment.package-operator.run/cache` etc.
If we ever refactor code so multiple finalizers are present on the same object, it's way easier to tell them apart and ensure that we are not removing the wrong finalizer.

internal/controllers/objectsetphases/phase_reconciler.go
thetechnick: Indeed. It's not obvious, that's why the comment is there.

internal/controllers/objectsetphases/phase_reconciler.go
thetechnick: It is correct - as far as I am able to tell, all the checks are being done in the `defaultAdoptionChecker`.
Only when `owner.GetStatusRevision()` is greater than current Revision is needAdoption set to true.

internal/controllers/objectsets/objectset_controller.go
thetechnick: To disable optimistic locking for this patch :)
Otherwise unrelated updates by other clients would return a ResourceVersion conflict error.

internal/controllers/objectsetphases/phase_reconciler_test.go
```diff
@@ -0,0 +1,743 @@
...
+	tests := []struct {
+		name          string
+		mockPrepare   func(*ownerStrategyMock, *phaseObjectOwnerMock)
+		object        client.Object
+		previous      []client.Object
+		errorAs       interface{}
+		needsAdoption bool
+	}{
+		{
+			name: "owned by newer revision",
```
garrettthomaskth, line 340:
```suggestion
			name: "owned by older revision",
```

internal/controllers/objectsetphases/phase_reconciler.go
```diff
@@ -0,0 +1,514 @@
...
+	}
+
+	currentRevision, err := getObjectRevision(obj)
+	if err != nil {
+		return false, fmt.Errorf("getting revision of object: %w", err)
+	}
+	if currentRevision > owner.GetStatusRevision() {
+		// no adoption, object is already owned by a newer revision.
+		return false, nil
+	}
```
garrettthomaskth, line 460: This is a duplicate of lines 433 to 440

internal/controllers/objectsetphases/phase_reconciler.go
```diff
@@ -0,0 +1,514 @@
...
+		}
+	}
+
+	// Object belongs to an older/lesser revision,
+	// is not already owned by us and also belongs to a previous revision.
+	return true, nil
+}
+
+func (c *defaultAdoptionChecker) isOwnedByPreviousRevision(
+	owner PhaseObjectOwner, obj client.Object, previous []client.Object,
```
garrettthomaskth, line 481:
```suggestion
func (c *defaultAdoptionChecker) isControlledByPreviousRevision(
	obj client.Object, previous []client.Object,
```

internal/ownerhandling/annotation_test.go
```diff
@@ -12,14 +12,40 @@ import (
...
+			Annotations: map[string]string{
+				ownerStrategyAnnotation: `[{"uid":"123456", "kind":"ConfigMap", "name":"cm1"}]`,
+			},
+		},
+	}
+	owner := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "cm1",
+			Namespace: obj.Namespace,
+			UID:       types.UID("123456"),
```
garrettthomaskth, line 37:
```suggestion
```

garrettthomaskth: I think all the UID's in this file can now be deleted :)

thetechnick: The current implementation is not needing them anymore, but they _should_ be present under normal conditions.
I would have just left them in there, because they don't hurt.

internal/controllers/controllers.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return nil
+}
+
+type dynamicWatchFreer interface {
+	Free(ctx context.Context, obj client.Object) error
+}
+
+// Frees caches and removes the associated finalizer.
+func FreeCacheAndFinalizer(
```
garrettthomaskth, line 62: Sorry, I think I already commented this but missed the answer. Why does cacheFinalizer have to be a parameter. Can't we just use the const string from objectset_controller.go?

internal/controllers/controllers.go
garrettthomaskth: Ohh okay that makes sense if there is a situation with two controllers adding the cache finalizer to one object, but I am kind of unclear what the point of that would be. Couldn't other issues also arise if there were two cache finalizers? From what I see, running the cacheFinalizer calls `FreeCacheAndFinalizer` which doesn't depend on the controller, only the object. So after the first cache finalizer is called the cache will already be freed, but it has a finalizer that says it needs to free the cache again.

internal/ownerhandling/annotation_test.go
garrettthomaskth: :+1:

internal/controllers/controllers.go
thetechnick: Freeing the cache multiple times is no problem.
But removing the finalizer before the cache is cleaned up is an issue, because if freeing the cache failed, we cannot add the finalizer again.

garrettthomaskth:
lgtm :)

internal/controllers/objectsetphases/phase_reconciler.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	// Set cache label
+	labels := desiredObj.GetLabels()
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[controllers.DynamicCacheLabel] = "True"
+	desiredObj.SetLabels(labels)
+
```
ashishmax31: Don't we also want to set the owner's revision in the desired object ?

thetechnick: :100: You are right, adding that now :+1:

thetechnick: fixed!

PR #27, opened 2022-09-15
implement ObjectSetPhase controller

thetechnick:


integration/objectset_test.go
```diff
@@ -25,385 +24,450 @@ import (
...
+			require.NoError(t,
+				Waiter.WaitForCondition(ctx, objectSetRev2, corev1alpha1.ObjectSetAvailable, metav1.ConditionFalse))
+			availableCond := meta.FindStatusCondition(objectSetRev2.Status.Conditions, corev1alpha1.ObjectSetAvailable)
+			require.NotNil(t, availableCond, "Available condition is expected to be reported")
+			assert.Equal(t, "ProbeFailure", availableCond.Reason)
+
+			// expect cm-1 to still be present and now owned by Rev2.
+			require.NoError(t, Client.Get(ctx, client.ObjectKey{
+				Name: cm1.Name, Namespace: objectSetRev1.Namespace}, currentCM1))
+			currentCM1.GetOwnerReferences()
```
garrettthomaskth, line 433:
```suggestion
			currentCM1.GetOwnerReferences()
assertIsController(t, objectSetRev2, currentCM1)
```

integration/objectset_test.go
```diff
@@ -25,385 +24,450 @@ import (
...
+			require.NoError(t,
+				Waiter.WaitForCondition(ctx, objectSetRev2, corev1alpha1.ObjectSetAvailable, metav1.ConditionTrue))
+
+			// Archive ObjectSet Rev1
+			require.NoError(t, Client.Patch(ctx, objectSetRev1,
+				client.RawPatch(types.MergePatchType, []byte(`{"spec":{"lifecycleState":"Archived"}}`))))
+			require.NoError(t,
+				Waiter.WaitForCondition(ctx, objectSetRev1, corev1alpha1.ObjectSetArchived, metav1.ConditionTrue))
+
+			// expect cm-2 to be gone.
```
garrettthomaskth, line 459: I think it makes sense for this to come directly after `// Expect ObjectSet Rev2 to become Available.`, because the availability of Rev2 is what is making cm-2 be cleaned up

thetechnick: yes but no. In the overall process that is orchestrated by the `ObjectDeployment` controller, latest revision available == True will immediately trigger all other pending revisions to get archived.
But `ObjectSets` themselves don't know about other `ObjectSets`, besides their predecessors.

This means that cm-2 will only be cleaned up _after_ rev1 is explicitly archived in the test run and only by using the higher-level API, you will have the full opinionated process in place.

We could pull more of that context into ObjectSet controller, but it would make coordination a lot more complicated and if we let it stay, other users can use ObjectSets directly but can create their own controller with different behavior and exit conditions than ObjectDeployment controller provides.

I'll change the comment in line 435 "// expect cm-2 to still be present. (will be cleaned up after Rev2 is Available)" to just `// expect cm-2 to still be present.`, sorry for adding to the confusion!

integration/objectset_test.go
thetechnick: Yeah, I was killing "assertIsController" because with phases, you cannot simply check if the object is owned by the ObjectSet anymore. It may also be owned by the phase object - which we don't even "see" from the test case.

I'll try to come up with a way to keep that check in.

PR #43, opened 2022-10-06
feat: package-loader

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <ykukreja@redhat.com>


cmd/package-loader/examples/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal@sha256:574f201d7ed185a9932c91cef5d397f5298dff9df08bc2ebb266c6d1e6284cd1
```
thetechnick, line 1:
```suggestion
FROM scratch
```
This is just a .tar holding a bunch of .yaml files with no executable no need to repackage a whole OS.
OLM is also using scratch as a base for these "data" images.

cmd/package-loader/examples/some-statefulset.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: some-stateful-set-1"
+  annotations:
+    package-operator.run/phase: "main-stuffe"
```
thetechnick, line 6:
```suggestion
    package-operator.run/phase: "main-stuff"
```

cmd/package-loader/examples/Dockerfile
yashvardhan-kukreja: So I considered that in the beginning but ultimately, we plan to use the image formed by the above Dockerfile as an initcontainer as a part of the unpackJob process.
And that process needs to run `cp` command to copy the contents of this image to the mounted `emptyDir` which will be shared by the package-loader container eventually.

Using scratch would fail that process because `cp` executable doesn't exist on `scratch` image.

Therefore, I proceeded to use the ubi-minimal image.

PR #44, opened 2022-10-06
introduce objectsetdeployment apis

ashishmax31:
Signed-off-by: ashish <asnaraya@redhat.com>


apis/core/v1alpha1/clusterobjectdeployment_types.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ClusterObjectDeploymentSpec defines the desired state of a ClusterObjectDeployment.
+type ClusterObjectDeploymentSpec struct {
+	// Number of old revisions in the form of archived ObjectSets to keep.
+	// +kubebuilder:default=5
+	RevisionHistoryLimit *int `json:"revisionHistoryLimit,omitempty"`
```
thetechnick, line 11: Kube is using `*int32` and default `10`, maybe we should do the same?
https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L370-L374

Please also change in ObjectDeploymentSpec!

apis/core/v1alpha1/clusterobjectdeployment_types.go
```diff
@@ -0,0 +1,59 @@
...
+	Template ObjectSetTemplate `json:"template"`
+}
+
+// ClusterObjectDeploymentStatus defines the observed state of a ClusterObjectDeployment
+type ClusterObjectDeploymentStatus struct {
+	// Conditions is a list of status conditions ths object is in.
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+	// DEPRECATED: This field is not part of any API contract
+	// it will go away as soon as kubectl can print conditions!
+	// Human readable status - please use .Conditions from code
```
thetechnick, line 24:
```suggestion
	// Phase is not part of any API contract
	// it will go away as soon as kubectl can print conditions!
	// When evaluating object state in code, use .Conditions instead.
```
Copy/pasta of new Phase comment from the other APIs

thetechnick (changes requested):
Thanks! Just 2 mini-nits.

ashishmax31:
Fixed @thetechnick, Have a look now.

thetechnick (approved):

PR #49, opened 2022-10-14
Introduce InTransition condition

thetechnick:
InTransition is set while an ObjectSet is not rolled out completely. This condition prevents ambeguity when looking at ObjectSet objects during an rollout process.
If an ObjectSet is InTransition, we will delay to report Succeeded until the rollout state is clean.

### DIFF
--- a/internal/controllers/objectsets/objectsetphases_reconciler.go
+++ b/internal/controllers/objectsets/objectsetphases_reconciler.go
@@ -82,18 +82,6 @@ func (r *objectSetPhasesReconciler) Reconcile(
 	}
 	objectSet.SetStatusControllerOf(controllerOf)
 
-	if !probingResult.IsZero() {
-		meta.SetStatusCondition(objectSet.GetConditions(), metav1.Condition{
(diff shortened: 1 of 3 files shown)


internal/controllers/objectsets/objectsetphases_reconciler.go
```diff
@@ -76,12 +76,25 @@ type phaseReconciler interface {
...
 ) (res ctrl.Result, err error) {
-	activeObjects, probingResult, err := r.reconcile(ctx, objectSet)
+	controllerOf, probingResult, err := r.reconcile(ctx, objectSet)
 	if err != nil {
 		return res, err
 	}
+	objectSet.SetStatusControllerOf(controllerOf)
+
+	inTransition := r.isObjectSetInTransition(objectSet, controllerOf)
+	if inTransition {
```
ashishmax31, line 86: Dont we also want to make sure that if an objectset is in transition we dont mark it as available?

ashishmax31 (approved):

PR #60, opened 2022-10-18
Remove ObjectDeployment revision annotation

thetechnick:
Remove ObjectDeploymentRevision annotation and rely on revision count reporting from ObjectSets.
Previously ObjectDeployment.metadata.generation was used to track the latest ObjectSet revision number, but .metadata.generation will diverge from the revision generation counter if fields outside the .spec.template.spec (e.g. RevisionHistoryLimit) are changed.

This PR is also closing a few different race conditions that come from using only inaccurate timestamps for revision ordering or not waiting for all ObjectSets to report a revision number.

### DIFF
--- a/internal/controllers/objectdeployments/new_revision_reconciler.go
+++ b/internal/controllers/objectdeployments/new_revision_reconciler.go
@@ -51,14 +51,6 @@ func (r *newRevisionReconciler) Reconcile(ctx context.Context,
 		return ctrl.Result{}, fmt.Errorf("errored when getting conflicting ObjectSet: %w", err)
 	}
 
-	// sanity check, before we increment the collision counter
-	conflictAnnotations := conflictingObjectSet.ClientObject().GetAnnotations()
-	currentDeploymentGeneration := fmt.Sprint(objectDeployment.ClientObject().GetGeneration())
-	if conflictAnnotations != nil && conflictAnnotations[DeploymentRevisionAnnotation] == currentDeploymentGeneration {
-		// Objectset for the current deployment revision already present, do nothing!
-		return ctrl.Result{}, nil
-	}
-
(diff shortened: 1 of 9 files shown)


internal/controllers/objectdeployments/new_revision_reconciler.go
```diff
@@ -51,14 +51,6 @@ func (r *newRevisionReconciler) Reconcile(ctx context.Context,
 		return ctrl.Result{}, fmt.Errorf("errored when getting conflicting ObjectSet: %w", err)
 	}
 
-	// sanity check, before we increment the collision counter
```
ashishmax31, line 54: Nico I believe there could be another race condition if we were to remove this sanity check. 
Let's say there are 2 revisions now, and then if two reconciliation loops were to start concurrently as part of the deployment hash changing. Now, both the reconciliation loops find that the revision for the current deployment version is not present and goes about trying to create a new revision.
Lets assume one creates the revision before the other, then the other reconciliation loop happening in the other go-routine would get an `IsAlreadyExists` error simply because the other go-routine was able to create the revision first. In this case there is no "actual" hash-collision but the code would go about incrementing the collision count.

thetechnick: Ashish, what you are describing is the exact scenario that the hash collision counter is enabling.
If two `ObjectDeployments` calculate the same hash for something (which can happen even with completely different contents), both need a way forward.

The case that you are describing seems perfectly safe, because if a Deployment decides to create a new revision and that hash is already taken, it needs to ++ it's collision counter or it will be stuck.

The same deployment is prevented from being reconciled by multiple goroutines by the shared workqueue within client-go/controller-runtime.

PR #61, opened 2022-10-19
feat: Package controller

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <ykukreja@redhat.com>


cmd/package-operator-manager/main.go
```diff
@@ -147,7 +148,7 @@ func run(log logr.Logger, scheme *runtime.Scheme, opts opts) error {
 	dc := dynamiccache.NewCache(
 		mgr.GetConfig(), mgr.GetScheme(), mgr.GetRESTMapper(), recorder,
 		dynamiccache.SelectorsByGVK{
-			// Only cache objects with our label selector,
+			// Only cache objects with our label selector,=
```
garrettthomaskth, line 151:
```suggestion
			// Only cache objects with our label selector,
```

config/images/package-loader.Containerfile
```diff
@@ -1,9 +1,9 @@
-FROM scratch
+FROM registry.access.redhat.com/ubi8/ubi-minimal@sha256:574f201d7ed185a9932c91cef5d397f5298dff9df08bc2ebb266c6d1e6284cd1
```
garrettthomaskth, line 1: Should this be done for all images in `config/`?

yashvardhan-kukreja: ohh whoops! I missed to revert this.

A few days back I used the ubi-minimal image for some blocker but that blocker sorted out and I forgot to revert it back to scratch.

Will do that :)

internal/controllers/controllers.go
```diff
@@ -22,16 +22,23 @@ const (
...
-// Ensures the given finalizer is set and persisted on the given object.
-func EnsureFinalizer(
+// Ensures the given finalizers are set and persisted on the given object.
+func EnsureFinalizers(
 	ctx context.Context, c client.Client,
-	obj client.Object, finalizer string,
+	obj client.Object, finalizers ...string,
 ) error {
-	if controllerutil.ContainsFinalizer(obj, finalizer) {
+	noFinalizersToPatch := true
```
garrettthomaskth, line 30:
```suggestion
	if len(finalizers) == 0 {
		return nil
	}
```

internal/controllers/controllers.go
```diff
@@ -49,17 +56,23 @@ func EnsureFinalizer(
...
-// Removes the given finalizer and persists the change.
-func RemoveFinalizer(
+// Removes the given finalizers and persist the change.
+func RemoveFinalizers(
 	ctx context.Context, c client.Client,
-	obj client.Object, finalizer string,
+	obj client.Object, finalizers ...string,
 ) error {
-	if !controllerutil.ContainsFinalizer(obj, finalizer) {
+	noFinalizersToPatch := true
```
garrettthomaskth, line 64:
```suggestion
	if len(finalizers) == 0 {
		return nil
	}
```

config/images/package-loader.Containerfile
yashvardhan-kukreja: So, the package-loader image requires command like `mkdir` and `cp` to be executed against it which aren't supported under a scratch container. Therefore, I proceeded with the ubi-minimal container.

yashvardhan-kukreja:
in favor of #67

PR #92, opened 2022-11-18
Mtsre 774 remote phase manager for every hosted cluster

garrettthomaskth:


thetechnick (changes requested):

internal/controllers/hostedclusters/adapter_hostedcluster.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type hostedCluster interface {
+	ClientObject() client.Object
+}
```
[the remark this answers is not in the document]
garrettthomaskth: >we only need that for other controllers that have a Cluster- and non Cluster- struct.
Ohh yeah, that is why I didn't put the `generic` prefix. 

I do think it would be nice to wrap the unstructured object. Are you saying that should be done in `hostedcluster_controller.go`?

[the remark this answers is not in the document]
thetechnick: Wrapping the unstructured object makes sense :)
But then this is not an `adapter_`

internal/controllers/hostedclusters/hostedcluster_controller.go
```diff
@@ -0,0 +1,108 @@
...
+	ctx = logr.NewContext(ctx, log)
+	hostedCluster := c.newHostedCluster()
+	if err := c.client.Get(ctx, req.NamespacedName, hostedCluster.ClientObject()); err != nil {
+		// Ignore not found errors on delete
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	desiredPackage := c.desiredPackage(hostedCluster)
+	// Can't use controllerutil.SetControllerReference because the HostedClusterType isn't in Scheme,
+	setControllerReference(hostedCluster.ClientObject(), desiredPackage)
```
thetechnick, line 53: `HostedCluster` does not need to be part of the scheme, controllerutil.SetControllerReference uses apituil.GVKForObject, which is interrogating the Unstructured object itself for Kind and GroupVersion :)
https://github.com/kubernetes-sigs/controller-runtime/blob/v0.13.1/pkg/client/apiutil/apimachinery.go#L76

garrettthomaskth: Ohhh awesome! Thanks! :)

internal/controllers/hostedclusters/adapter_hostedcluster.go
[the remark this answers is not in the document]
garrettthomaskth: Ahhhh :D Okay thanks!

thetechnick:
@garrettthomaskth Why are you replacing the project logos? :D

garrettthomaskth:
Hey Nico! Sorry that was the precommit hooks!

thetechnick:
/hold

Needs integration tests

cmd/package-operator-manager/main.go
```diff
@@ -71,6 +74,9 @@ func main() {
 	flag.StringVar(&opts.loadPackage, "load-package", "", "(internal) runs the package-loader sub-component to load a package mounted at /package")
 	flag.StringVar(&opts.selfBootstrap, "self-bootstrap", "",
 		"(internal) bootstraps Package Operator with Package Operator using the given Package Operator Package Image")
+	flag.StringVar(&opts.remotePhasePackageImage, "remote-phase-package-image", os.Getenv("PKO_REMOTE_PHASE_PACKAGE_IMAGE"),
+		"Providing an image will enabling the deployment of an additional controller that creates a remote phase manager package"+
```
garrettthomaskth, line 78:
```suggestion
		"Providing an image will enable the deployment of an additional controller that creates a remote phase manager package"+
```

cmd/package-operator-manager/main.go
```diff
@@ -71,6 +74,9 @@ func main() {
...
+			"for every HostedCluster object")
```
garrettthomaskth, line 79: Either "Image to use for the remote phase manager package" or "Image to use for the remote phase manager package that is created for every HostedCluster object if HyperShift HostedCluster API is present on the cluster"

integration/hypershift_test.go
```diff
@@ -0,0 +1,48 @@
...
+	hc := &hypershiftv1alpha1.HostedCluster{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "test-hc",
+			Namespace: "default",
+		},
+	}
+	err := Client.Create(ctx, hc)
+	require.NoError(t, err)
+	defer cleanupOnSuccess(ctx, t, hc)
+
```
garrettthomaskth, line 31: Is there a way to check here that no package was created?

magefile.go
```diff
@@ -126,6 +128,7 @@ func init() {
 
 	// Extra dependencies must be specified here to avoid a circular dependency.
 	packageImages[pkoPackageName].ExtraDeps = []interface{}{Generate.PackageOperatorPackage}
+	packageImages[remotePhasePackageName].ExtraDeps = []interface{}{Generate.RemotePhasePackage}
```
garrettthomaskth, line 131: Lines 64-66 say "// Note that you can't reference the Generate mage target in ExtraDeps
	// since that would result in a circulat dependency. They must be added via init() for now." If that comment is out of date could you remove it?

magefile.go
```diff
@@ -342,13 +352,18 @@ func patchPackageOperatorManager(scheme *k8sruntime.Scheme, obj *unstructured.Un
...
 
-	var packageOperatorManagerImage string
+	var (
+		packageOperatorManagerImage string
+		remotePhasePackageImage     string
+	)
 	if len(os.Getenv("USE_DIGESTS")) > 0 {
 		// To use digests the image needs to be pushed to a registry first.
-		mg.Deps(mg.F(Build.PushImage, "package-operator-manager"))
+		mg.Deps(mg.F(Build.PushImage, "package-operator-manager", remotePhasePackageName))
```
garrettthomaskth, line 361: I think `PushImage` only takes one argument. Maybe this is what you mean `mg.Deps(mg.F(Build.PushImage, "package-operator-manager"), mg.F(Build.PushImage, pkoPackageName))`?

magefile.go
```diff
@@ -342,13 +352,18 @@ func patchPackageOperatorManager(scheme *k8sruntime.Scheme, obj *unstructured.Un
...
 		packageOperatorManagerImage = locations.ImageURL("package-operator-manager", true)
+		remotePhasePackageImage = locations.ImageURL(remotePhasePackageName, true)
 	} else {
 		packageOperatorManagerImage = locations.ImageURL("package-operator-manager", false)
```
garrettthomaskth, line 365: Should running `Dev.LoadImage` on the images be a dependency for the case where we don't use digests?

integration/hypershift_test.go
thetechnick: Negative assertions are not that nice,  because we can only set them up time-based.
E.g. in 10s no Package was created, as we are having this completely covered with unittests, just testing the success end-to-end test is ok.

magefile.go
thetechnick: Right

magefile.go
thetechnick: I want all of this to be refactored and either moved into devkube or the CLI.
Especially the digest handling needs to be moved into the CLI, because I am not sure it works as intended and is very brittle...

I don't think loading images in here is necessary.

magefile.go
thetechnick: Let's revisit the whole magefile next year, I think most of the comments are still alright.

thetechnick (approved):

PR #96, opened 2022-11-22
Add build subcommand to kubectl-package.

eqrx:
This commands builds a source directory into a container image and pushes it to a given destination.


internal/cli/build/build.go
```diff
@@ -0,0 +1,75 @@
...
+	if err := crane.Push(p.Image, destination, opts...); err != nil {
+		return fmt.Errorf("push: %w", err)
+	}
+
+	return nil
+}
+
+// Build the package described by sourcePath.
+func Build(sourcePath string) (*Package, error) {
+	rawLayer := map[string][]byte{}
```
thetechnick, line 36: Important: please put all files into a /package directory.
I've tried it top-level at first as well, but all the linux directories like /etc /dev and sysfs are cluttering root and make loading difficult.

thetechnick:
@eqrx what do you think about the way kubectl is structuring it's commands:
E.g. https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/taint/taint.go

It separates cobra configuration code from options and their validation. I quite like it, because testing becomes very straightforward as it's not as intertwined with cobra.

eqrx:
@thetechnick I see the benefit but I am not sure if this would e a benefit right now since the complexity of our commands is relatively low. I cut down the complexity of the build subcommand in the last push to 25 lines, ptal. Would you still say it would be beneficial now?

Right now I would argue that pulling the command definition and the validation apart makes it harder to read

eqrx:
nvm, have implemented something like it :D

is it close enough for kubectl for you or do you want what they do in verbatim?

thetechnick (changes requested):

thetechnick (approved):

PR #97, opened 2022-11-28
internal/packages refactoring

thetechnick:
Refactor the `internal/packages` package to make the loading process more modular.
This is in preparation of adding PKO package validation to the CLI and reports error reporting into the API as a side effect.

Example of new error output containing detailed error locations and YAML document index within that document
```yaml
  - lastTransitionTime: "2022-11-29T14:33:17Z"
    message: |
      Package validation errors:
      - Missing package-operator.run/phase Annotation in manifest.yaml#0
    observedGeneration: 1
    reason: LoadError
    status: "True"
    type: Invalid
```


internal/packages/packagebytes/transform.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		fileMap[path] = content
+	}
+
+	return nil
+}
+
+func (t *TemplateTransformer) transform(ctx context.Context, path string, content []byte) ([]byte, error) {
+	if !packages.IsYAMLFile(path) || packages.IsManifestFile(path) {
+		// not a YAML file, skip
```
garrettthomaskth, line 64:
```suggestion
		// It is the package manifest file or not a YAML file, skipping 
```

internal/packages/packagebytes/validate.go
```diff
@@ -0,0 +1,19 @@
+package packagebytes
+
+import "context"
+
+type Validator interface {
+	Validate(ctx context.Context, fileMap FileMap) error
+}
+
```
garrettthomaskth, line 8:
```suggestion
var _ Validator = (ValidatorList)(nil)

```

internal/packages/packagestructure/content.go
```diff
@@ -0,0 +1,71 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"sigs.k8s.io/yaml"
+
+	corev1alpha1 "package-operator.run/apis/core/v1alpha1"
+	manifestsv1alpha1 "package-operator.run/apis/manifests/v1alpha1"
+	"package-operator.run/package-operator/internal/packages"
+	"package-operator.run/package-operator/internal/packages/packagebytes"
+)
+
+// PackageContent represents a the parsed content of an PKO package.
```
garrettthomaskth, line 15:
```suggestion
// PackageContent represents the parsed content of an PKO package.
```

internal/packages/packagestructure/content.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (pc *PackageContent) ToTemplateSpec() (templateSpec corev1alpha1.ObjectSetTemplateSpec) {
+	templateSpec.AvailabilityProbes = pc.PackageManifest.Spec.AvailabilityProbes
+
+	objectsByPhase := map[string][]unstructured.Unstructured{}
+	for _, objects := range pc.Manifests {
+		for _, object := range objects {
+			annotations := object.GetAnnotations()
+			phase := annotations[manifestsv1alpha1.PackagePhaseAnnotation]
```
garrettthomaskth, line 48: Why do we want to remove the phase annotations from the object?

internal/packages/packagestructure/content.go
```diff
@@ -0,0 +1,92 @@
...
+			delete(annotations, manifestsv1alpha1.PackagePhaseAnnotation)
+			if len(annotations) == 0 {
+				// This is important!
+				// When submitted to the API server empty maps will be dropped.
+				// Semantic equality checking is considering a nil map to ne not equal to an empty map.
+				// And if semantic equality checking fails, hash collision checks will always find a hash collision if the ObjectSlice already exists.
+				annotations = nil
+			}
+			object.SetAnnotations(annotations)
+			objectsByPhase[phase] = append(objectsByPhase[phase], object)
```
garrettthomaskth, line 58:
```suggestion
			objectsByPhase[phase] = append(objectsByPhase[phase], corev1alpha1.ObjectSetObject{
				Object: object,
			}))
```

internal/packages/packagestructure/content.go
```diff
@@ -0,0 +1,92 @@
...
+		return nil, fmt.Errorf("marshal YAML: %w", err)
+	}
+	fm[packages.PackageManifestFile] = packageManifestBytes
+	return fm, nil
+}
+
+func (pc *PackageContent) ToTemplateSpec() (templateSpec corev1alpha1.ObjectSetTemplateSpec) {
+	templateSpec.AvailabilityProbes = pc.PackageManifest.Spec.AvailabilityProbes
+
+	objectsByPhase := map[string][]unstructured.Unstructured{}
```
garrettthomaskth, line 44:
```suggestion
	objectsByPhase := map[string][]corev1alpha1.ObjectSetObject{}
```

internal/packages/packagestructure/content.go
```diff
@@ -0,0 +1,92 @@
...
+		phase := corev1alpha1.ObjectSetTemplatePhase{
+			Name:  phase.Name,
+			Class: phase.Class,
+		}
+		unstructuredObjects := objectsByPhase[phase.Name]
+		for _, obj := range unstructuredObjects {
+			phase.Objects = append(phase.Objects, corev1alpha1.ObjectSetObject{
+				Object: obj,
+			})
+		}
```
garrettthomaskth, line 72:
```suggestion
		phase.Objects = objectsByPhase[phase.Name]
```

internal/packages/packagestructure/content.go
thetechnick: I wanted to remove this annotation because it's only significant during package "assembly" and I don't see a reason for persisting it into the cluster.

internal/packages/packagestructure/content.go
thetechnick: Good find, this makes the code easier to read and more efficient :)

internal/packages/packagestructure/content.go
garrettthomaskth: I guess I am wondering if it could ever be useful to have that information there while debugging on the cluster or during an incident

internal/packages/packagestructure/transform.go
```diff
@@ -0,0 +1,86 @@
...
+		base = map[string]string{}
+	}
+	for k, v := range additional {
+		base[k] = v
+	}
+	if len(base) == 0 {
+		return nil
+	}
+	return base
+}
```
garrettthomaskth, line 86: can be replaced with `utils.mergeKeysFrom`

internal/packages/packagebytes/transform.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"text/template"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"package-operator.run/package-operator/internal/packages"
+)
+
+type Transformer interface {
+	Transform(ctx context.Context, fileMap FileMap) error
```
eqrx, line 15: This enforces that the FileMap will be modified in place. Unsure if that is good or a Signature like
`Transform(ctx context.Context, fileMap FileMap) (FileMap, error)` would be nicer to use for some applications

thetechnick: This would require copying the FileMap over and over again. As this map can be a few MB in size, I would rather not clone it.

eqrx (approved):

PR #99, opened 2022-11-28
Add build subcommand to kubectl-package.

eqrx:
Rewrite kubectl-package so the version subcommand is more maintanable. Build command allows to pack a package source directory into an image. Leaving stubs to allows integration of verification into this later. Optionally it is possible to push to image or dump it into a tar.gz file.


thetechnick (changes requested):
Please run hack/ensure-and-run-goimports.sh on all new files to format imports.

thetechnick (approved):

PR #107, opened 2022-12-01
Cli binary attached to releases

eqrx:
Let GH release action attach cli multi arch binaries to the created release.

This requires mage to name binaries as how they are supposed to show up in the release.

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -44,10 +44,27 @@ const (
 	golangciLintVersion  = "1.50.1"
 	kindVersion          = "0.16.0"
 	k8sDocGenVersion     = "0.5.1"
+
+	cliCmdName = "kubectl-package"
 )
 
+type archTarget struct {
+	OS, Arch string
+}
+
+type command struct {
+	ReleaseArchitectures []archTarget
+}
+
 var (
-	cliExtraArch = [][2]string{{"linux", "amd64"}, {"darwin", "amd64"}, {"darwin", "arm64"}}
+	commands = map[string]command{
+		"package-operator-manager": {nil},
+		"remote-phase-manager":     {nil},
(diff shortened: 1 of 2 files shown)


thetechnick:
Can we keep these files in the `bin/` directory?
The bin directory is excluded from git, having these files top-level makes it very easy to check in a binary...

```
$ git status
On branch multiarch
Your branch is up to date with 'eqrx/multiarch'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        kubectl-package.darwin.amd64
        kubectl-package.darwin.arm64
        kubectl-package.linux.amd64

nothing added to commit but untracked files present (use "git add" to track)
```

magefile.go
```diff
@@ -391,20 +420,28 @@ type Build mg.Namespace
...
+	for name, cmd := range commands {
+		for _, arch := range cmd.ReleaseArchitectures {
+			targets = append(targets, mg.F(Builder.Cmd, name, arch.OS, arch.Arch))
+		}
+	}
+	mg.Deps(targets...)
+
+	for name, cmd := range commands {
+		for _, arch := range cmd.ReleaseArchitectures {
+			must(sh.Copy(fmt.Sprintf("%s.%s.%s", name, arch.OS, arch.Arch), binaryPath(name, arch)))
```
thetechnick, line 442:
```suggestion
			must(sh.Copy(fmt.Sprintf("%s_%s_%s", name, arch.OS, arch.Arch), binaryPath(name, arch)))
```

Isn't underscore more common than separating by dots?
e.g. binaries here: https://github.com/kcp-dev/kcp/releases/tag/v0.9.1

eqrx: since we already use "-" within the command name it is not a good one. "kubectl-package-linux-amd64".

But if you want that I can implement it

eqrx:
I had problems with our gh release action when release files were not on top level. but maybe that was just something else. will change this

thetechnick (approved):

PR #109, opened 2022-12-01
Move all the path section in magefile into one place.

eqrx:


thetechnick (changes requested):
I like your changes to restructure our path handling, but we have made a conscious decision to have everything work just within the project checkout directory.

Reason is to not have any dangling stuff lying around on a users computer when they remove the project.

Instead of caching in user-home by default, can we stick to defaulting to within the project checkout directory and have an env-var as override to store it somewhere else?

eqrx:
(done)

thetechnick (approved):

PR #111, opened 2022-12-01
Add extra validators for labels and GVK

thetechnick:

### DIFF
--- a/cmd/kubectl-package/command/build.go
+++ b/cmd/kubectl-package/command/build.go
@@ -75,7 +75,11 @@ func (b Build) Run(ctx context.Context) error {
 
 	verboseLog.Info("validating package image")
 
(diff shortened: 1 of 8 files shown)


cmd/kubectl-package/command/build.go
```diff
@@ -75,7 +75,11 @@ func (b Build) Run(ctx context.Context) error {
 
 	verboseLog.Info("validating package image")
 
-	structureLoaderOpts := []packagestructure.LoaderOption{packagestructure.WithManifestValidators(&packagestructure.ObjectPhaseAnnotationValidator{})}
+	structureLoaderOpts := []packagestructure.LoaderOption{
+		packagestructure.WithManifestValidators(
+			packagestructure.DefaultValidators,
```
eqrx, line 80: Since there are now three users of this piece of code (build subcmd, validate subcmd and mage pkg builder in my PR) ... maybe we could have a "gimme leader with default validator" shortcut for this?

eqrx (changes requested):

eqrx (approved):

PR #112, opened 2022-12-01
Package test framework

thetechnick:
Introduce test framework for package templates
    
Templating yaml, while being very flexible, can be error prone and 
impact of changes is often hard to judge.
    
This PR introduces a testing framework for validating package
templating results to improve confidence and opens the door to handle more
complex configuration inputs or topology dependend application layout.
    
Template test cases are a list of template inputs that will be executed
 to check if the expected output is produced. After executing the
template, all other structural checks are also run.
    

.pre-commit-config.yaml
```diff
@@ -15,6 +15,7 @@ repos:
   - id: check-symlinks
   - id: detect-private-key
   - id: end-of-file-fixer
+    exclude: ".test-fixtures\/.*"
```
eqrx, line 18: why?:)

eqrx: won't this also be detected with golangci?

also, why not let the fixer do its job and check in the corrected form instead of bypassing it?

eqrx (approved):

PR #113, opened 2022-12-05
Refactor cli

eqrx:


eqrx:
(adding the label myself since review already passed)

PR #131, opened 2023-01-16
Refactor internal/packages

eqrx:


internal/packages/packageexport/image.go
```diff
@@ -0,0 +1,65 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+
+	"package-operator.run/package-operator/internal/packages"
+	"package-operator.run/package-operator/internal/packages/packagecontent"
+)
+
+func Image(files packagecontent.Files) (v1.Image, error) {
+	configFile := &v1.ConfigFile{Architecture: "amd64", OS: "linux", Config: v1.Config{}, RootFS: v1.RootFS{Type: "layers"}}
```
thetechnick, line 19: Can you please add a comment, why we are hardcoding Architecture and OS for the moment?

I'll look into opening tickets to re-engineer our in-cluster package loading process so we can drop this.

Comment suggestion:
// Hardcoded to linux/amd64 or kubernetes will refuse to pull the image on our target architecture.
// We will drop this after refactoring our in-cluster package loading process to make it architecture agnostic.

thetechnick (changes requested):
LGTM, one small nit where we should add a comment to the code.

thetechnick (approved):

PR #140, opened 2023-01-27
Replace image tags with digests on package build phase

kostola:


kostola:
Closing this because we changed the approach

PR #141, opened 2023-01-27
Add "update" subcommand

kostola:


apis/manifests/v1alpha1/packagemanifest_types.go
```diff
@@ -135,6 +146,28 @@ func (m *PackageManifest) Validate() error {
 		}
 	}
 
+	specImages := field.NewPath("spec").Child("images")
+	if len(m.Spec.Images) > 0 {
```
eqrx, line 150: Nit: You could remove the condition for this block and and always iterate over `m.Spec.Images`. Downside would be that the allocation of `existingNames` would sometimes be unnecessary but I would argue that is this a rather miniscule price to reduce the cyclo complexity here.

If you do not know yet, a nil slice is considered empty, you can iterate over it.

apis/manifests/v1alpha1/packagemanifest_types.go
```diff
@@ -135,6 +146,28 @@ func (m *PackageManifest) Validate() error {
...
+		existingNames := []string{}
+		for i, image := range m.Spec.Images {
+			if len(image.Name) < 1 {
```
eqrx, line 153: Nit: A switch case could look better here:)

apis/manifests/v1alpha1/packagemanifestlock_types.go
```diff
@@ -0,0 +1,71 @@
...
+	Digest string `json:"digest"`
+}
+
+// Validates the PackageManifestLock and returns an aggregated list of all errors.
+func (m *PackageManifestLock) Validate() error {
+
+	var allErrs field.ErrorList
+
+	specImages := field.NewPath("spec").Child("images")
+	if len(m.Spec.Images) > 0 {
```
eqrx, line 38: See above regarding removing the condition and the switch

apis/manifests/v1alpha1/packagemanifestlock_types.go
```diff
@@ -0,0 +1,71 @@
...
+		}
+	}
+
+	if len(allErrs) == 0 {
+		return nil
+	}
+	return allErrs.ToAggregate()
+}
+
+func init() {
```
eqrx, line 69: Nit: I personally would put the init function right below the imports since it can cause a lot of headache if overlooked (speaking from experience :D)

cmd/kubectl-package/command/updatecmd/updatecmd.go
```diff
@@ -0,0 +1,85 @@
...
+	case args[0] == "":
+		return fmt.Errorf("%w: target path empty", cmdutil.ErrInvalidArgs)
+	}
+
+	u.Target = args[0]
+	return nil
+}
+
+func (u Update) Run(ctx context.Context, out io.Writer) (err error) {
+	var (
```
eqrx, line 40: Nit: `var filemap packagecontent.Files`

eqrx (commented):
Didn't you say you are new to go? Does not look like it, good job :) 

Have written down a couple of nits, (really) feel free to ignore them, jut though I throw some in :)

apis/manifests/v1alpha1/packagemanifest_types.go
kostola: Thx for the suggestion! Changed :+1:

apis/manifests/v1alpha1/packagemanifestlock_types.go
kostola: Thx for the suggestion! Changed :+1:

apis/manifests/v1alpha1/packagemanifestlock_types.go
kostola: I placed it here as in `packagemanifest_types.go` ...if we decide to move we should do it everywhere imho

cmd/kubectl-package/command/updatecmd/updatecmd.go
kostola: Thx for the suggestion! Changed :+1:

apis/manifests/v1alpha1/packagemanifestlock_types.go
```diff
@@ -0,0 +1,61 @@
...
+
+	for name, image := range thisImages {
+		otherImage, exists := otherImages[name]
+		if !exists || otherImage != image {
+			return false
+		}
+	}
+
+	return true
+}
```
thetechnick, line 57: Can you please move this code into the updatecmd package?
I would like to keep as little business logic as possible in public API package, so we don't have to include it in our consistency guarantees.

kostola: Done :+1:

PR #148, opened 2023-02-03
Make to gosec linter less obnoxious by disabling some checks

eqrx:


kostola (approved):
I am not an expert of gosec linter but this seems good :+1: 

I have no idea however if there are more rules to exclude, but we'll figure it out along the way.

PR #149, opened 2023-02-03
Update deps in magefile and CI

eqrx:


kostola (approved):
@eqrx I trust you on this one :sweat_smile: :+1:

PR #152, opened 2023-02-06
Add manifest.lock.yaml image and digest validation in build subcommand

kostola:


cmd/kubectl-package/command/buildcmd/build.go
```diff
@@ -108,3 +116,53 @@ func (b *Build) CobraCommand() *cobra.Command {
 
 	return cmd
 }
+
+func preBuildValidation(pkg *packagecontent.Package) error {
+	if pkg.PackageManifestLock == nil {
+		return fmt.Errorf("manifest.lock.yaml is missing (try running \"kubectl package update\")")
```
eqrx, line 122: `return fmt.Errorf(`manifest.lock.yaml is missing (try running "kubectl package update")`)`
also works if you prefer it

eqrx (approved):

PR #160, opened 2023-02-07
Allow cli build to push images to local container runtime

eqrx:


eqrx:
Closing since it seems not very useful right now

PR #170, opened 2023-02-13
Object Template Controller

garrettthomaskth:


cmd/package-operator-manager/main.go
```diff
@@ -227,6 +228,7 @@ func runLoader(scheme *runtime.Scheme, packageKey client.ObjectKey) error {
 	return nil
 }
 
+//nolint:maintidx
```
eqrx, line 231: Lazy, lazy :D

PR #181, opened 2023-02-20
Add command line arguments to kubectl-package tree to support config

eqrx:


robshelly (approved):
:+1: Verified that the above solution works using the PKO template examples:
```
$ ../package-operator/bin/kubectl-package tree 1_applications/2_templates/ --config-path 1_applications/2_templates/.test-fixtures/.config.yaml 
nginx
Package namespace/name
└── Phase deploy
    └── apps/v1, Kind=Deployment /name
```

PR #220, opened 2023-03-01
Improve go dependency update management

eqrx:
- Virtually not limit dependabot PR count
- Do not track go.work.sum


eqrx:
Don't mind me, testing around with prow

PR #237, opened 2023-03-14
Pko advanced self config

thetechnick:
Enable PKO to configure itself to support setting affinity and toleration rules, possible also requests/limits next.
Also setups a local development registry to enable us to "digestify" our image references instead of relying on tags, by using the new `kubectl package update` command.


Nanyte25:
@thetechnick happy to try test these changes if your waiting on others?

thetechnick:
@Nanyte25 It would be super helpful, if you could run `./mage dev:deploy` on your local machine to validate everything is working.
Still finishing unittest coverage and final UX for injecting configuration from the self-bootstrap job.

kostola (approved):
Based on my current knowledge, this looks good to me

PR #238, opened 2023-03-17
[MTSRE-962] feat: allow specification of externally declared for status probing

Ajpantuso:
### Summary

Adds support to probe external objects, those not directly declared by the package, for status within individual phases.


thetechnick:
@Ajpantuso Really nice PR!
Just pushed the hash collision test, every time we are changing the lower-level APIs, the hash changes and we have to update it manually right now...

I have no reservations merging this change in, after we got some integration tests setup.

In a follow up PR, we also have to add this to the Packaging logic, so phases[].externalObjects is populated from the Package filesystem.
Maybe by introducing a `package-operator.run/external: True` annotation for objects that are only getting watched but not created?
Code ref: https://github.com/package-operator/package-operator/blob/main/internal/packages/packagecontent/totemplate.go#L10

Ajpantuso:
> @Ajpantuso Really nice PR! Just pushed the hash collision test, every time we are changing the lower-level APIs, the hash changes and we have to update it manually right now...
> 
> I have no reservations merging this change in, after we got some integration tests setup.
> 
> In a follow up PR, we also have to add this to the Packaging logic, so phases[].externalObjects is populated from the Package filesystem. Maybe by introducing a `package-operator.run/external: True` annotation for objects that are only getting watched but not created? Code ref: https://github.com/package-operator/package-operator/blob/main/internal/packages/packagecontent/totemplate.go#L10

Added the logic to apply external objects from the filesystem. Will play around with integration tests (and probably unit tests for the affected plugin commands) this coming week.

thetechnick (approved):

PR #240, opened 2023-03-21
enforce JSONPath notation

garrettthomaskth:


garrettthomaskth:
> Nice PR! `JSONPathFormatError` needs to be included here: https://github.com/garrettthomaskth/package-operator/blob/prefix_sources_with_dots/internal/controllers/objecttemplate/template_reconciler.go#L342 Or it won't be reported to the user in the condition.

Thanks! I thought every error that is returned by `copySourceItems` would be wrapped in a `SourceError`  [here](https://github.com/garrettthomaskth/package-operator/blob/prefix_sources_with_dots/internal/controllers/objecttemplate/template_reconciler.go#L136) and then because it is now a `SourceError` it would be reported to the user in the condition. Or am I misunderstanding the code?

thetechnick (approved):

PR #274, opened 2023-04-26
[MTSRE-1170] refactor: overhaul of kubectl-package

Ajpantuso:
### Summary

Refactoring `kubectl-package` command structure in preparation of future feature development.


internal/cmd/validate_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/stretchr/testify/require"
+	"gotest.tools/v3/assert"
+
+	"package-operator.run/package-operator/internal/packages/packagecontent"
+)
+
+func TestValidateConfig(t *testing.T) {
+	t.Parallel()
+
+	testLogger := funcr.New(func(_, _ string) {}, funcr.Options{})
```
eqrx, line 20: Use https://github.com/go-logr/logr/blob/v1.2.4/testr/testr.go#L29

eqrx (changes requested):
Hey, thank you very much for the pr! I ran over it and found that one issue so faR. Since this PR is a bit bigger I would continue reviewing it after our sprint, is that alright with you?:)

internal/cmd/validate_test.go
Ajpantuso: Sure, this can wait, but the longer it sits the more rebases will be required. So for review purposes I'm looking more for feedback on whether the commands still work the way they are intended to. The actual refactor should be fairly straight forward.

eqrx: Right you are, let me do that now

cmd/kubectl-package/build/build.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	internalcmd "package-operator.run/package-operator/internal/cmd"
```
eqrx, line 11: The reason for the original naming is to avoid named imports. k8s is full of this and my favorite example is the following:

```
	schemacell "k8s.io/apiextensions-apiserver/pkg/apiserver/schema/cel"
	pkgapiscell "k8s.io/apiserver/pkg/apis/cel"
	pkgcell "k8s.io/apiserver/pkg/cel"
```

gopls and a lot of IDEs have trouble to autoresolve dependencies when every package is basically named the same so I am always trying to give it a less ambiguous name.

cmd/kubectl-package/tree/tree.go
```diff
@@ -0,0 +1,86 @@
...
+		configPathUse     = "file containing config which is used for templating."
+	)
+
+	flags.BoolVar(
+		&o.ClusterScope,
+		"cluster",
+		o.ClusterScope,
+		clusterScopeUse,
+	)
+	flags.StringVar(
```
eqrx, line 74: I like this

internal/cmd/build.go
```diff
@@ -0,0 +1,190 @@
...
+	pkgLockImages := map[string]v1alpha1.PackageManifestLockImage{}
+	for _, image := range pkg.PackageManifestLock.Spec.Images {
+		pkgLockImages[image.Name] = image
+	}
+
+	// check that all the images in manifest file exists in lock files too, and their "image" fields are the same
+	for imageName, image := range pkgImages {
+		lockImage, existsInLock := pkgLockImages[imageName]
+		if !existsInLock {
+			return err("image \"%s\" exists in manifest but not in lock file (try running \"kubectl package update\")", imageName)
```
eqrx, line 134: Would recommend to use
```
"a \"b\""
```

```
`a "b"`
```

Ajpantuso:
Waiting on #281 for rebase

internal/cmd/update.go
```diff
@@ -0,0 +1,188 @@
...
+	pkg, err := packageloader.New(scheme).FromFiles(ctx, fileMap)
+	if err != nil {
+		return nil, err
+	}
+
+	return pkg, nil
+}
+
+type Clock interface {
+	Now() v1.Time
```
eqrx, line 181: Nit: Most interfaces that describe clocks let `Now` return `time.Time`. Should be use a distinguished function name like `MetaTime` here to future proof?

cmd/kubectl-package/treecmd/tree_test.go
```diff
@@ -16,12 +14,12 @@ func TestTree_Success(t *testing.T) {
 	t.Run("namespace scoped", func(t *testing.T) {
 		t.Parallel()
 
-		cmd := command.CobraRoot()
+		cmd := NewCmd()
 		stdout := &bytes.Buffer{}
 		stderr := &bytes.Buffer{}
 		cmd.SetOut(stdout)
 		cmd.SetErr(stderr)
-		cmd.SetArgs([]string{"tree", "--config-testcase", "namespace-scope", "testdata"})
```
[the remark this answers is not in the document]
Ajpantuso: The idea was to isolate these packages and then add real `integration` for `kubectl-package` later which will actually build the binary and run these commands as a user would. The integration suite would another 1k lines to this PR so trying to limit scope at the moment.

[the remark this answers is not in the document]
eqrx: Agreeable, resolving!

internal/cmd/update.go
Ajpantuso: For an internal embedded interface I don't think it's important. `v1.Time` acts just like `time.Time` in this context.

eqrx: oh right, overlooked that

eqrx (approved):

internal/cmd/validate_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+			cfg.Option(tc.Options...)
+
+			tc.Assertion(t, cfg.Validate())
+		})
+	}
+}
+
+//go:embed testdata/manifest.yaml
+var _manifest []byte
```
eqrx, line 103: Why is this prefixed with an underscore?

eqrx (approved):

internal/cmd/validate_test.go
Ajpantuso: I follow this advice https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_. Helps me avoid mistakes when using global state, but totally up to preference.

PR #277, opened 2023-04-28
Let the tree subcommanduse the first testcase for config if nothing specified

eqrx:


cmd/kubectl-package/command/treecmd/tree.go
```diff
@@ -91,6 +91,11 @@ func (t *Tree) Run(ctx context.Context, out io.Writer) error {
 			return fmt.Errorf("%w: test template with name %s not found", cmdutil.ErrInvalidArgs, t.ConfigTestcase)
 		}
 	default:
+		rawConfig := pkg.PackageManifest.Test.Template[0].Context.Config.Raw
+
+		if err := json.Unmarshal(rawConfig, &config); err != nil {
+			return fmt.Errorf("unmarshal config from test template %s: %w", t.ConfigTestcase, err)
```
Ajpantuso, line 97: This looks like a copy/paste message. Can it be update to remove the `t.ConfigTestCase` option?

Ajpantuso (changes requested):

cmd/kubectl-package/command/treecmd/tree.go
eqrx: The `t.ConfigTestCase` branch picks a testcase with the given name while the default one picks the first one when no name is specified. So no, not really copy paste

eqrx: oh right, my bad

PR #282, opened 2023-05-02
[MTSRE-1171] feat: improve initialization for kubectl-package

Ajpantuso:
### Summary

- Sets up uniform initialization for subcommands so that the kubect-package root can inject dependencies
- Sets up `zap` as the logger for `kubectl-package` and adds zap options to the persistent flags
- Repackages the root command under `kubectl-package`

### Change Type

New Feature

### Check List Before Merging


cmd/kubectl-package/deps/command.go
```diff
@@ -0,0 +1,127 @@
...
+type defaultRendererFactory struct {
+	logFactory LogFactory
+	scheme     *runtime.Scheme
+}
+
+func (f *defaultRendererFactory) Renderer() treecmd.Renderer {
+	return internalcmd.NewTree(
+		f.scheme,
+		internalcmd.WithLog{
+			Log: f.logFactory.Logger().V(1),
```
eqrx, line 61: Is all output of tree debug output? (zaps DebugLevel lvl)

Ajpantuso: I'm fine removing the `v(1)`'s for this and the `builder`. It's a hold over from when we weren't using levels. I've set `zap` to default to error so `Info` level will not be shown unless the user asks for it.

Ajpantuso: The `V(1)`'s are gone now.

eqrx (approved):

PR #291, opened 2023-05-09
[MTSRE-1201] Add 'insecure' option and tests using local image registry

Ajpantuso:
### Summary

Adds `--insecure` flag to relevant `kubectl-package` subcommands and includes integration testcases which push/pull from image registries.

### Change Type

New Feature

### Check List Before Merging


integration/kubectl-package/testdata/images/invalid-package.tar
eqrx, line 1: its not really practical  to build those on demand & automatically when the test is executed, right?

eqrx (changes requested):
As always, thanks for the good work!

eqrx (approved):

PR #304, opened 2023-05-22
Expose environment information to template engine

thetechnick:
### Summary
Exposes environment information to the template engine to allow package authors to react to different Kubernetes Versions, detect whether they are running on OpenShift and use global Proxy configuration in OpenShift.

### Change Type
New Feature 

### Check List Before Merging


cmd/package-operator-manager/main.go
```diff
@@ -105,22 +107,40 @@ type packageOperatorManager struct {
...
 	if err := allControllers.SetupWithManager(mgr); err != nil {
 		return nil, err
 	}
+
+	var envSinks []environment.Sink
+	for _, c := range allControllers.List() {
+		envSink, ok := c.(environment.Sink)
+		if !ok {
+			envSinks = append(envSinks, envSink)
+		}
```
Ajpantuso, line 128:
```suggestion
		if ok {
			envSinks = append(envSinks, envSink)
		}
```

internal/environment/environment.go
```diff
@@ -0,0 +1,161 @@
...
+const (
+	environmentProbeInterval    = 1 * time.Minute
+	openShiftClusterVersionName = "version"
+	openShiftProxyName          = "cluster"
+)
+
+type Sink interface {
+	InjectEnvironment(env manifestsv1alpha1.PackageEnvironment)
+}
+
```
Ajpantuso, line 30:
```suggestion
type Sinker interface {
	SetEnvironment(env manifestsv1alpha1.PackageEnvironment)
}

type Sink struct {
    env  manifestsv1alpha1.PackageEnvironment
    lock sync.RWMutex
}

func (s *Sink) SetEnvironment(env manifestsv1alpha1.PackageEnvironment) {
	s.lock.Lock()
	defer s.lock.Unlock()
	s.env = *env.DeepCopy()
}

func (s *Sink) GetEnvironment() manifestsv1alpha1.PackageEnvironment {
    s.lock.RLock()
	defer s.lock.RUnlock()
	return *s.env.DeepCopy()
}
```

Very optional, but embedding an implementation of this interface into reconcilers probably makes sense as the behavior would just be duplicated otherwise.

internal/environment/environment.go
```diff
@@ -0,0 +1,161 @@
...
+	return nil
+}
+
+func (m *Manager) do(ctx context.Context) error {
+	env, err := m.probe(ctx)
+	if err != nil {
+		return err
+	}
+	for _, s := range m.sinks {
+		s.InjectEnvironment(env)
```
Ajpantuso, line 71: Using the `Sink` implementation mentioned above would help, but as written these calls both block reconciliation and are blocked by reconciliation. With one reconciler it's  not a real concern, but it might make sense to use go routines and a waitgroup to reduce contention if more reconcilers use this.

Ajpantuso (changes requested):
One issue, but other comments are just nits really. Overall looks okay.

PR #323, opened 2023-06-09
Do not override CGO_ENABLED if env is already set when building binaries

eqrx:
### Check List Before Merging


Nanyte25 (approved):
/lgtm

PR #324, opened 2023-06-21
[MTSRE-1277] Fix garbage collection of archived revisions

robshelly:
### Summary
Currently archived `objectsets` are not being garbage collected according to the `revisionHistoryLimit`. This PR fix this bug so when new revisions are created archived ones are deleted accordingly. `objectsets` are also configured to only store the same number of `previousRevisions` to avoid 'not found' errors when looking up previous revisoins which have been garbage collected. 

### Change Type

 Bug Fix 

### Check List Before Merging

### DIFF
--- a/internal/controllers/objectdeployments/archive_reconciler.go
+++ b/internal/controllers/objectdeployments/archive_reconciler.go
@@ -225,9 +225,13 @@ func (a *archiveReconciler) garbageCollectRevisions(ctx context.Context, previou
 	if deploymentRevisionLimit != nil {
 		revisionLimit = *deploymentRevisionLimit
 	}
-	numToDelete := len(previousObjectSets) - (int(revisionLimit))
-	for numToDelete > 0 {
-		if err := a.client.Delete(ctx, previousObjectSets[0].ClientObject()); err != nil {
+	numToDelete := len(previousObjectSets) - int(revisionLimit)
+	for _, previousObjectSet := range previousObjectSets {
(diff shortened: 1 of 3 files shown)


internal/controllers/objectdeployments/archive_reconciler.go
```diff
@@ -226,3 +218,20 @@ func intersection(a, b []objectIdentifier) (res []objectIdentifier) {
...
+	if deploymentRevisionLimit != nil {
+		revisionLimit = *deploymentRevisionLimit
+	}
+	numToDelete := len(previousObjectSets) - (int(revisionLimit))
+	for numToDelete > 0 {
+		if err := a.client.Delete(ctx, previousObjectSets[0].ClientObject()); err != nil {
+			return fmt.Errorf("failed to delete objectset: %w", err)
+		}
+		numToDelete--
+	}
```
thetechnick, line 234:
```suggestion
	numToDelete := len(previousObjectSets) - int(revisionLimit)
	for _, previousObjectSet := range previousObjectSets {
		if numToDelete <= 0 {
			break
		}

		if err := a.client.Delete(ctx, previousObjectSet.ClientObject()); err != nil && !errors.IsNotFound(err) {
			return fmt.Errorf("failed to delete objectset: %w", err)
		}
		numToDelete--
	}
```

- The () arround the int-cast are not needed
- The for loop needs to delete _all_ Archived ObjectSets that are above the history limit

PR #329, opened 2023-06-27
Update dependencies that do not break controller-runtime

eqrx:
### Check List Before Merging


.gitignore
```diff
@@ -18,3 +18,6 @@ bin
 
 # JetBrains IDEs (e.g. GoLand)
 .idea/
+
+# Autogenerated files
+go.work.sum
```
erdii, line 23: This should already be in main. (Is it a no-op to just leave this here?)

integration/package-operator/objecttemplate_test.go
```diff
@@ -44,7 +44,7 @@ func TestObjectTemplate_creationDeletion_packages(t *testing.T) {
 
 	kubernetesKey := "kubernetes"
 	kubernetesPath := ".environment.kubernetes.version"
-	kubernetesValue := "v1.27.1"
+	kubernetesValue := "v1.27.3"
```
erdii, line 47: Sooo... I don't exactly remember, but did we find out why this change needs to happen?

eqrx: because the int test has a specific k8s version hard coded and fixing this is not part of the PR :|

eqrx: Issue created

erdii: Alrighty... thankersens!

PR #344, opened 2023-07-12
Let unit tests run in parallel

eqrx:
### Summary

This change does
- add nolint directives to all integration tests
- add t.Parallel calls to all non integration tests
- prevents parallel tests from using loop variables

### Check List Before Merging

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -27,7 +27,6 @@ linters:
   #- nakedret         # Naked return good return.
   #- nlreturn         # Could be nice in the future.
   #- nonamedreturns   # Named return good return.
-  #- paralleltest     # Would be nice but we can't run all the integration tests in parallel and there is not a way to disable the linter for paths.
   #- testpackage      # Could be nice in the future.
   #- unparam          # Not that useful/false positives.
   #- varnamelen       # Not that useful/false positives.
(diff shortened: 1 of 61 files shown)


.golangci.yaml
```diff
@@ -77,6 +76,7 @@ linters:
   - noctx
   - nolintlint
   - nosprintfhostport
+  - paralleltest
```
erdii, line 79: How about this?

```suggestion
  - paralleltest # Parallel tests are forbidden in our integration tests. This linter can be disabled on a per-package basis for other tests.
```

.golangci.yaml
erdii, line 1: We can add this exlude-rule so we don't need the `nolint:paralleltest` comments in the integration tests.

```yaml
issues:
  exclude-rules:
    # Integration tests MUST NOT run in parallel.
    - path: '(.+)_test\.go'
      linters:
      - paralleltest
```

WDYT about adding this exclude rule and then dropping all the `nolint:paralleltest` mentions from `./integration/*`?

internal/controllers/phase_reconciler_test.go
erdii, line 1: Line 1239 has a for loop by value which then captures the value inside the loop scope like this: `tc := tc`
I think this should be reformatted to use an index accessor, like you did with all the other for loops.
WDYT?

PR #345, opened 2023-07-19
Bootstrap syncs registryHostOverrides config with existing PKO deployment

kostola:
### Check List Before Merging


integration/package-operator/zzz_upgrade_test.go
```diff
@@ -0,0 +1,213 @@
...
+	log.Info("Installing latest released PKO")
+	require.NoError(t, createAndWaitFromHTTP(ctx, []string{LatestSelfBootstrapJobURL}))
+
+	jobList := &batchv1.JobList{}
+	require.NoError(t, Client.List(
+		ctx, jobList,
+		client.InNamespace("package-operator-system"),
+	))
+	for i := range jobList.Items {
+		require.NoError(t, Waiter.WaitToBeGone(ctx, &jobList.Items[i], func(obj client.Object) (done bool, err error) { return false, nil }, dev.WithTimeout(6*time.Minute)))
```
erdii, line 55: Maybe we should factor this magic timeout at the start of this file?

integration/package-operator/environment.go
```diff
@@ -60,6 +62,8 @@ func init() {
 		panic(err)
 	}
 
+	LatestSelfBootstrapJobURL = os.Getenv("PKO_TEST_LATEST_BOOTSTRAP_JOB")
```
erdii, line 65: I think we should log this url somewhere in the tests to make debugging easier

integration/package-operator/zzz_upgrade_test.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	require.NoError(t, Waiter.WaitForCondition(ctx, pkg, corev1alpha1.PackageAvailable, metav1.ConditionTrue, dev.WithTimeout(5*time.Minute)))
+	log.Info("Latest released PKO is now available")
+
+	cosList := &corev1alpha1.ClusterObjectSetList{}
+	require.NoError(t, Client.List(ctx, cosList, client.MatchingLabels{"package-operator.run/object-deployment": "package-operator"}))
```
erdii, line 62: I think we have variable definitions for the label name (and maybe value too?) somewhere in the codebase

integration/package-operator/zzz_upgrade_test.go
```diff
@@ -0,0 +1,213 @@
...
+	require.Len(t, cosList.Items, 1)
+	existingCos := &cosList.Items[0]
+
+	log.Info("Apply self-bootstrap-job.yaml built from sources")
+	require.NoError(t, createAndWaitFromFiles(ctx, []string{filepath.Join("..", "..", "config", "self-bootstrap-job.yaml")}))
+
+	jobList = &batchv1.JobList{}
+	require.NoError(t, Client.List(
+		ctx, jobList,
+		client.InNamespace("package-operator-system"),
```
erdii, line 72: this hardcoded namespace won't work in openshift, will it?

integration/package-operator/zzz_upgrade_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func deleteExistingPKO(ctx context.Context) error {
+	log := logr.FromContextOrDiscard(ctx)
+
+	cosList := &corev1alpha1.ClusterObjectSetList{}
+	err := Client.List(
+		ctx, cosList,
+		client.MatchingLabels{
+			"package-operator.run/object-deployment": "package-operator",
```
erdii, line 90: see comment above

integration/package-operator/zzz_upgrade_test.go
```diff
@@ -0,0 +1,213 @@
...
+		},
+	)
+	if err != nil {
+		return err
+	}
+	for i := range cosList.Items {
+		cos := &cosList.Items[i]
+		log.Info(fmt.Sprintf("Removing finalizers from cluster object set %s: %v", cos.Name, cos.Finalizers))
+		patch := fmt.Sprintf("{\"metadata\":{\"resourceVersion\":\"%s\",\"finalizers\":[]}}", cos.ResourceVersion)
```
erdii, line 99: could be written as a raw string literal to make the patch more readable for humans and thus less error prone:

```suggestion
		patch := fmt.Sprintf(`{"metadata":{"resourceVersion":"%s","finalizers":[]}}`, cos.ResourceVersion)
```

magefile.go
```diff
@@ -716,11 +725,6 @@ func (Test) packageOperatorIntegration(ctx context.Context, filter string) {
 	}
 	args = append(args, "./integration/package-operator/...")
 
-	_, isCI := os.LookupEnv("CI")
```
eqrx, line 719: Is that supposed to be removed?

eqrx (changes requested):

magefile.go
erdii: Yup. This was accidentally added with the intent to have code coverage on int tests but it doesn't do anything.

PR #346, opened 2023-07-24
[MTSRE-1338] Consider resource scope in isObjectSetInTransition check

kostola:
### Summary

This PR tries to fix the scenario described below.

If a `Package` contains a mix of namespaced resources and cluster resource, the package gets stuck in `Progressing` state when installed because:

- `LatestRevisionPendingSuccess` condition is never removed
- to have the `LatestRevisionPendingSuccess` condition be removed, the corresponding `ObjectSet` must enter `Succeeded` state
- the underlying `ObjectSet` never reaches `Succeeded` because the `InTransition` condition never goes away
- the `InTransition` condition never gets removed because the underlying `isObjectSetInTransition` function in the code [assumes that all the resources controlled by a `Package` are part of its namespace](https://github.com/package-operator/package-operator/blob/9165fcc405c745acd8af85e805128ab346793b7a/internal/controllers/objectsets/objectsetphases_reconciler.go#L287-L289), and it fails on cluster scoped resources

### Change Type

Bug Fix

### Check List Before Merging

### DIFF
--- a/cmd/package-operator-manager/components/objectset.go
+++ b/cmd/package-operator-manager/components/objectset.go
@@ -2,6 +2,7 @@ package components
 
 import (
 	"github.com/go-logr/logr"
+	"k8s.io/client-go/discovery"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"package-operator.run/package-operator/internal/controllers/objectsets"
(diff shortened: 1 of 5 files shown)


thetechnick:
Superseded by:
https://github.com/package-operator/package-operator/pull/360

PR #349, opened 2023-08-01
Refactor internal/packages

eqrx:
### Check List Before Merging


eqrx:
Int tests fail on CI .... assuming the problem is coming from main

eqrx:
This PR need go1.20. We need to be ubi compatible so I will keep this PR open until november since the support is supposed to come then.

PR #352, opened 2023-08-02
Enable Template helpers

thetechnick:
### Summary
Allow users to define template blocks to reuse template blocks.

Also stops the test-fixture generator to create empty files, which becomes annoying for helper files just containing define blocks.

### Change Type
New Feature

### Check List Before Merging


docs/api-reference.md
```diff
@@ -29,6 +29,7 @@ apiVersion: package-operator.run/v1alpha1
 kind: ClusterObjectDeployment
 metadata:
   name: example
+  namespace: default
```
eqrx, line 32: this file is not part of the PR

eqrx (changes requested):

docs/api-reference.md
erdii: fix: rebase onto main and redo the commit, k8s-docgen has been fixed

eqrx (approved):

PR #353, opened 2023-08-02
.gitignore go.work.sum since it is autogenerated anyway

eqrx:
### Check List Before Merging


erdii:
Clear guidelines from docs on `go.work.sum` and checking it into vcs don't seem to exist [1]. Since it's annoying that this file changes all the time without obvious reason is annoying. it doesn't seem to be needed for working dev and builds and other big projects gitignore the file as well [2], I built and tested locally and CI passes, I'll say:

/lgtm
/approve

[1] https://github.com/golang/go/issues/51941
[2] https://github.com/github/gitignore/pull/4081

PR #401, opened 2023-08-16
Retract package-operator.run/package-operator

eqrx:
### Check List Before Merging


eqrx:
we need to double check if that is actually working before merging :D

eqrx:
You can retract versions, you can't retract the whole module :(

PR #428, opened 2023-08-25
MTSRE-1454 fix HTTP(S) proxy configuration coming from in-cluster objects

erdii:
Go caches proxy settings that are read from the envvars `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY. This means that our bootstrap procedure:
1. detect if running in kube or openshift
2. (in openshift) detect if cluster proxy settings are configured and override bootstrap process proxy environment variables
3. assume that from now on all http/https calls that are not excluded via `NO_PROXY` will go through the configured proxy

will not work because the calls to the kubernetes api have already read and cached the envvars in go code.

The workaround is:
- lookup proxy settings from cluster and environment variables
- if they are different: merge proxy settings into environment variables and replace current pko process with a new instance by using unix.Exec (execve)

### Change Type

Bug Fix

### Check List Before Merging

### DIFF
--- a/cmd/package-operator-manager/bootstrap/bootstrap.go
+++ b/cmd/package-operator-manager/bootstrap/bootstrap.go
@@ -6,10 +6,12 @@ import (
 	"os"
 
 	"github.com/go-logr/logr"
+	"golang.org/x/sys/unix"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"package-operator.run/cmd/package-operator-manager/bootstrap/proxy"
 	"package-operator.run/cmd/package-operator-manager/components"
 	"package-operator.run/internal/controllers"
 	"package-operator.run/internal/environment"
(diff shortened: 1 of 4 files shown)


erdii:
I'm still working on unit and integration tests for this change.
@thetechnick and me had the idea of locking down the kind cluster (by disabling nat) so that we are forced to always develop and test with a proxy.

Locking down would be hackishly done with:
`$CONTAINER_COMMAND exec -it package-operator-dev-control-plane iptables -t nat -I POSTROUTING 3 -j RETURN`
which basically rejects all pod traffic that wants to leave the cluster and thus must be NATed.
Even though this is a hacky workaround this worked fine and **could** be added as a helper to devkube but has one big gap: Restarting the kind node container makes it "forget" the injected iptables rule. This means that a dev with a running kind cluster can reboot their machine, restart the cluster by running `podman start package-operator-dev-control-plane` and continue working, without realising that the pods in the cluster now have direct internet access.

Other things that have to be solved:
- [x] tricking pko into thinking that the kind cluster is an openshift cluster with active proxy settings (need to enable and create 2 OpenShift CRs `ClusterVersion` and `Proxy`
- [x] deploy a proxy with `hostNetwork: true`
- [ ] deploy a second dev registry that is only reachable by the proxy (easy: let it run as a sidecar in the proxy pod and inject a custom hostname for 127.0.0.1 into the proxy pod - this hostname is now usable for talking to the registry through the proxy)
- [ ] refactor magefile to enable pushing test package images into the proxy-only registry
- [ ] add integration tests that ensure that installing packages through the proxy-only registry is working correctly

erdii:
Team decision: get the fix out - work on the proxy-first dev/integration environment separately

thetechnick (approved):

PR #427, opened 2023-08-25
make (default)ImageOrg in magefile truly configurable

erdii:
### Check List Before Merging

- [x] This PR passes all pre-commit hook validations.
- [ ] This PR is fully tested and regression tests are included.
- [x] Relevant documentation has been updated.

### Additional Information

test using 
```bash
./mage dev:teardown
IMAGE_ORG=eef-freef.io/sbubby ./mage dev:deploy

export KUBECONFIG=$PWD/.cache/dev-env/kubeconfig.yaml
kubectl get deployment -n package-operator-system package-operator-manager -o yaml \
  | yq '.spec.template.spec.containers[0].image'
# should output something like:
# eeef-freef.io/sbubby/package-operator-manager@sha256:e8d3ed5dd738dc7820f5a05bc8b4ae083686efd85827f7f821ae11f9272ba9c3
```

PR #430, opened 2023-08-25
retroactively apply PR review suggestion from @kostola

erdii:
This is in a new PR because prow merged the other PR without waiting for the change request to be closed: https://github.com/package-operator/package-operator/pull/427#discussion_r1305396387


kostola (approved):
So minimalist yet so beautiful

PR #467, opened 2023-09-08
[MTSRE-1471] Fix race condition in objectset handover integration tests

erdii:
by moving  assertions that should happen after the new object set becomes available after the code that waits for it to become available.

Previous to this fix, testing a cpu-starved PKO would often result in a failure when the tests asserted that object had already fully been handed over, when pko only had reported that the new object set was already created and unavailable due to failing probes on an already existing object.

```
    hypershift_test.go:91:
        	Error Trace:	/home/erdii/projects/redhat/github.com/package-operator/package-operator/integration/package-operator/objectset_test.go:763
        	           				/home/erdii/projects/redhat/github.com/package-operator/package-operator/integration/package-operator/objectset_test.go:505
        	           				/home/erdii/projects/redhat/github.com/package-operator/package-operator/integration/package-operator/hypershift_test.go:91
        	Error:      	Should be true
        	Test:       	TestHyperShift/ObjectSetHandover
        	Messages:   	controller name of default-test-hc/cm-1 not prefixed with test-rev2
--- FAIL: TestHyperShift (74.36s)
    --- PASS: TestHyperShift/HyperShift_Installation (1.11s)
    --- PASS: TestHyperShift/ObjectSetSetupPauseTeardown (17.16s)
    --- FAIL: TestHyperShift/ObjectSetHandover (2.04s)
FAIL
```

PS: why does `go mod tidy` bump a dependency here?

### Change Type

Bug Fix

### Check List Before Merging

### Additional Information

Reproducing the bug can by done by patching the pko deployment template file and removing pko package test-fixtures prior to running the integration tests on main.
Validating the fix can be done by then checking out the fix (while keeping the patched deployment template), and rerunning integration tests (with a cluster teardown in between).

```bash
# removing package test fixtures
rm -rf packages/package-operator/.test-fixtures

# patching the deployment template:
git apply <(cat <<EOF
diff --git a/config/packages/package-operator/package-operator-manager.Deployment.yaml.gotmpl b/config/packages/package-operator/package-operator-manager.Deployment.yaml.gotmpl
index 72bc06b..6ff9b6e 100644
--- a/config/packages/package-operator/package-operator-manager.Deployment.yaml.gotmpl
+++ b/config/packages/package-operator/package-operator-manager.Deployment.yaml.gotmpl
@@ -93,10 +93,10 @@ spec:
         # default resources
         resources:
           limits:
-            cpu: 200m
+            cpu: 10m
             memory: 400Mi
           requests:
-            cpu: 200m
+            cpu: 10m
             memory: 300Mi
 {{- end}}
 {{- if hasKey . "environment" }}
EOF
)

# recreating cluster and running integration tests:
./mage dev:teardown dev:integration
```

### DIFF
--- a/integration/package-operator/objectset_test.go
+++ b/integration/package-operator/objectset_test.go
@@ -418,6 +418,15 @@ func defaultObjectSetRev2(cm1, cm3 *corev1.ConfigMap, rev1 *corev1alpha1.ObjectS
 	}, nil
 }
 
+func isControllerOf(objectReferences []corev1alpha1.ControlledObjectReference, expected corev1alpha1.ControlledObjectReference) bool {
+	for _, ref := range objectReferences {
+		if reflect.DeepEqual(ref, expected) {
+			return true
+		}
+	}
+	return false
+}
+
 func runObjectSetHandoverTest(t *testing.T, namespace, class string) {
 	t.Helper()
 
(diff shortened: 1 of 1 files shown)


integration/package-operator/objectset_test.go
```diff
@@ -418,6 +418,15 @@ func defaultObjectSetRev2(cm1, cm3 *corev1.ConfigMap, rev1 *corev1alpha1.ObjectS
 	}, nil
 }
 
+func isControllerOf(objectReferences []corev1alpha1.ControlledObjectReference, expected corev1alpha1.ControlledObjectReference) bool {
```
eqrx, line 421: I could have sworn we had this method flying around somewhere but I can't find it anymore. Would it be a good idea to move this one here to some util pkg?

integration/package-operator/objectset_test.go
```diff
@@ -418,6 +418,15 @@ func defaultObjectSetRev2(cm1, cm3 *corev1.ConfigMap, rev1 *corev1alpha1.ObjectS
...
+	for _, ref := range objectReferences {
+		if reflect.DeepEqual(ref, expected) {
```
eqrx, line 423: Isn't there a nicer way than using reflect here? :D I mean, in k8s it seems to be usual but I can't help it

erdii:
Superseded by https://github.com/package-operator/package-operator/pull/495

PR #502, opened 2023-09-21
Let bootstrapper only handle package operator packages

eqrx:
### Check List Before Merging


thetechnick (changes requested):
This PR mus also stop the ClusterObjectSet and ClusterObjectDeployment controllers from working on other packages during bootstrap.

internal/controllers/objectdeployments/objectdeployment_controller.go
```diff
@@ -106,25 +106,25 @@ func NewClusterObjectDeploymentController(
...
-func (od *GenericObjectDeploymentController) Reconcile(
-	ctx context.Context, req ctrl.Request,
-) (ctrl.Result, error) {
+func (od *GenericObjectDeploymentController) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, err error) {
 	log := od.log.WithValues("ObjectDeployment", req.String())
 	defer log.Info("reconciled")
+
+	/*if od.isBootstrap && req.Name != "package-operator" {
+		return
+	}*/
```
erdii, line 119: Since we have multi-component packages coming, we'll probably split the package-operator package into separate packages (for example putting the webhooks into a sub-component).

This will obviously not work with a single-hardcoded name, so I think we should use a labelSelector to select PKO components that the bootstrapper will reconcile.

PR #505, opened 2023-09-22
Add file access from template engine

thetechnick:
### Summary
Allows to include static - non templated - files into templates.

### Change Type
New Feature

### Check List Before Merging

- [x This PR passes all pre-commit hook validations.


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/disiqueira/gotree v1.0.0
 	github.com/go-logr/logr v1.2.4
+	github.com/gobwas/glob v0.2.3
```
eqrx, line 12: https://pkg.go.dev/path/filepath#Glob can't be used since you want to use `**` right

internal/packages/packagecontent/topackage.go
```diff
@@ -15,9 +17,11 @@ func PackageFromFiles(ctx context.Context, scheme *runtime.Scheme, files Files)
 	pkg = &Package{nil, nil, map[string][]unstructured.Unstructured{}}
 	for path, content := range files {
 		switch {
-		case !packages.IsYAMLFile(path):
+		case !packages.IsYAMLFile(path) ||
+			strings.HasPrefix(filepath.Base(path), "_"):
```
eqrx, line 21:
```suggestion
		case strings.HasPrefix(filepath.Base(path), "_"):
			// skip template helper files.
		case !packages.IsYAMLFile(path):
```

eqrx (changes requested):

go.mod
thetechnick: yeah, I've checked stdlib and think that the import makes sense.

eqrx (approved):

PR #545, opened 2023-10-13
Make probing a public package.

thetechnick:
### Summary
Moves the `package-operator.run/internal/probing` package to be public under `package-operator.run/pkg/probing`.

### Change Type

Refactoring

### Check List Before Merging


go.mod
```diff
@@ -4,13 +4,14 @@ go 1.20
 
 replace package-operator.run/apis => ./apis
 
+replace package-operator.run/pkg => ./pkg
```
eqrx, line 7: suggest does not work for some reason: 
```
replace (
package-operator.run/apis => ./apis
package-operator.run/pkg => ./pkg
)
```

internal/controllers/objectsetphases/objectsetphase_reconciler.go
```diff
@@ -15,7 +15,8 @@ import (
 
 	corev1alpha1 "package-operator.run/apis/core/v1alpha1"
 	"package-operator.run/internal/controllers"
-	"package-operator.run/internal/probing"
+	internalprobing "package-operator.run/internal/probing"
```
eqrx, line 18: I suppose we still have some code that we do not want to push to the submodule since we do not want to guarantee stability, right?

pkg/probing/fieldsequal.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// FieldsEqualProbe checks if the values of the fields under the given json paths are equal.
```
eqrx, line 11: Do we want to enable docs linters for the probing package? I get that we do not want it in the main module but it could be nice here

eqrx (changes requested):

go.mod
erdii, line 7:
```suggestion
replace (
	package-operator.run/apis => ./apis
	package-operator.run/pkg => ./pkg
```

eqrx: josh managed to add that as an actual suggestion block :D

eqrx (approved):

PR #553, opened 2023-10-18
MTSRE-1571: Fix YAML document split breaks PEM certs

thetechnick:
### Summary
When loading YAML files, we did blindly split on ---, so if someone added a PEM encoded certificate that is using ------ add start/end, we where splitting the document...

This commit changes the behavior to only split if the --- is in it's own line and there are no preceding characters.

### Change Type
Bug Fix

### Check List Before Merging


erdii:
/lgtm

Using a decoder would be nicer though but kube's yaml lib does not have one.

PR #546, opened 2023-10-13
Correctly handle config fields in multi component packages

kostola:
### Summary

Multi-component packages were breaking whenever the root package or one of its components contained `spec.config` in its `manifest.yaml` file.

This PR aims to fix the issue.

### Check List Before Merging

- [ ] This PR passes all pre-commit hook validations.
- [ ] This PR is fully tested and regression tests are included.
- [ ] Relevant documentation has been updated.

PR #568, opened 2023-10-19
Correctly handle config fields in multi component packages

kostola:
### Summary

**IMPORTANT: this PR supersedes #546**

Multi-component packages were breaking whenever the root package or one of its components contained `spec.config` in its `manifest.yaml` file.

This PR aims to fix the issue.

### Check List Before Merging


internal/cmd/tree.go
```diff
@@ -72,7 +72,13 @@ func (t *Tree) RenderPackage(ctx context.Context, srcPath string, opts ...Render
 		return "", fmt.Errorf("loading package contents from folder: %w", err)
 	}
 
-	pkg, err := packagecontent.PackageFromFiles(ctx, t.scheme, files, cfg.Component)
+	// TODO: show all components in the tree
```
thetechnick, line 75: Instead of listing EVERYTHING with tree, wouldn't it make more sense if we add a `--component` flag to the tree command to only render the component in question?
At least that would be more what I would expect to happen.

thetechnick (approved):

internal/cmd/tree.go
kostola: will do in a following PR 👍

PR #570, opened 2023-10-19
Improve our extensive assorted collection of linters 🎩

eqrx:
### Check List Before Merging


.golangci.yaml
```diff
@@ -97,6 +100,7 @@ linters:
   - usestdlibvars
   - wastedassign
   - whitespace
+  - zerologlint
```
erdii, line 103:
```suggestion
```

PR #592, opened 2023-10-23
Replace interface{} usage with any

eqrx:
Replace all occurreces of `interface{}` with `any`. `any` is an alias from `interface{}` and defines more clearly that you want to use "anything" instead of "specifically something that does not need to do anything though"


erdii:
This PR does not pass linting.

robshelly:
:+1: Simple change just replacing interface{} with any but verified anyway by building and deploying some packages.

robshelly (approved):

erdii (dismissed):

thetechnick (approved):

PR #648, opened 2023-11-12
Update pre-commit modules

eqrx:


.pre-commit-config.yaml
```diff
@@ -22,11 +22,6 @@ repos:
     - --autofix
   - id: trailing-whitespace
 
-- repo: https://github.com/dnephin/pre-commit-golang
-  rev: v0.3.5
-  hooks:
-  - id: go-fmt
```
thetechnick, line 28: Why is this removed?
Is it now included with the version upgrade?

eqrx: Hey @thetechnick, gofumpt is already a part of golangci-lint, it uses a superset of what gofmt enforces.

thetechnick (approved):

PR #690, opened 2023-11-17
Add depguard to prevent importing x/exp/(maps|slices)

eqrx:


eqrx:
Noticed that the exp packages are still used for development and contain useful features. Closing.

PR #764, opened 2023-12-19
Make kubectl package intgration test appropriately

eqrx:


thetechnick:
@eqrx Why?

eqrx:
@thetechnick 
this way you can just call `go test ./...` to run unit tests and `go test --tag=integration` if you want to run integration tests.
If int tests are not tagged you must either:
- See that calling `go test ./...` just explodes
- You need to manually search for what folder contains what kind of tests
- Figure out which mage command does #2 :D
I agree that it is possible to just run the mage target but I find it more straight forward to allow users to just call go test like they expect it

PR #795, opened 2024-01-09
Use cardboard instead of mage

eqrx:


kostola (approved):
I can reliably use this instead of mage for my local developent. LGTM.

PR #805, opened 2024-01-11
Hypershift environment integration

thetechnick:
### Summary
Adds information about HyperShift to the environment data that is exposed to Package and ObjectTemplate.

### Change Type
New Feature

### Check List Before Merging


internal/controllers/objecttemplate/template_reconciler.go
```diff
@@ -304,18 +306,23 @@ func (r *templateReconciler) templateObject(
...
+
+	envData := map[string]any{}
+	packageEnvironment, err := json.Marshal(env)
 	if err != nil {
-		return env, fmt.Errorf("marshaling: %w", err)
+		return envData, fmt.Errorf("marshaling: %w", err)
 	}
 
-	err = json.Unmarshal(packageEnvironment, &env)
+	err = json.Unmarshal(packageEnvironment, &envData)
```
erdii, line 321: This looks expensive. What is the rationale behind remarshalling the environment data to/from json here?

erdii (commented):
I would like to have a unit test showcasing the templating features implemented here and also to serve as a regression test when iterating on the features.

I'd volunteer to implement the unit test, too.

internal/controllers/objecttemplate/template_reconciler.go
erdii: From a voice chat with @thetechnick:

The rationale is: Values are exposed to the package templates through maps instead of go values so that template authors can use lower-case field names to access data (like this: `.foo.bar.baz`) which mixes better with the surrounding yaml manifests of a template.


Picture this:

```yaml
# ...
spec:
  lbCount: {{ .Config.Loadbalancers.Count }}
```

vs this:

```yaml
# ...
spec:
  lbCount: {{ .config.loadbalancers.count }}
```

The second example feels more intuitive to write and reason about, especially if you also write the manifest the provides the accessed data and then are confused about specifying the data in lowercase field names and accessing the data with uppsercase names on the other side.

The json remarshal roundtrip into a `map[string]any` is a quick and dirty implementation of this.

internal/environment/environment.go
```diff
@@ -220,8 +266,38 @@ func (s *Sink) SetEnvironment(env *manifests.PackageEnvironment) {
...
+
+	// Lookup HostedCluster
+	hostedClusterList := &hypershiftv1beta1.HostedClusterList{}
+	if err := s.client.List(ctx, hostedClusterList); err != nil {
+		return nil, fmt.Errorf("listing HostedClusters: %w", err)
+	}
+	for _, hc := range hostedClusterList.Items {
+		hcNamespace := hypershiftv1beta1.HostedClusterNamespace(hc)
+		if hcNamespace != namespace {
+			continue
```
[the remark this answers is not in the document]
erdii: Update: the string transformation done by `hypershiftv1beta1.HostedClusterNamespace(hc)` is not reversible, so there is no way to compute the HostedCluster name from the cluster namespace name. We also haven't seen management clusters with more than several hundred HostedCluster objects, PKO has all of them in cache and we can optimize this later by adding a custom search index if needed.

erdii (approved):

PR #861, opened 2024-01-25
Extend package validation to components in the same run

kostola:
### Summary

Currently running `kubectl package validate` for a multi component package only validates the root component. In order to validate all the structure, a subsequent call for each component must be made.

With this PR, calling the command passing the root folder will trigger recursive validation of the components too.

### Change Type

New Feature

### Check List Before Merging


eqrx (changes requested):
just looking at the codecov report. do you think it is possible to increase the coverage a bit?:)

thetechnick:
Looks good, but extra coverage would be really appreciated :)

eqrx (approved):

PR #930, opened 2024-02-16
W.I.P to add image load feature to pko cli.

Nanyte25:
### Summary
I am trying to create a load command for the Kubectl package CLI tool

To load Taz images from a file location in the local runtime registry.

### Check List Before Merging


Nanyte25:
@eqrx and @kostola I would appreciate your feedback thanks

.DS_Store
kostola, line 1: @Nanyte25 you committed this by mistake

kostola:
@Nanyte25 this PR adds a new dedicated subcommand, but as far as I understood the initial idea was to have this feature available via a flag of the `build` subcommand. The logic seems good, it just needs to be moved.

Is this correct @thetechnick @eqrx ?

PR #951, opened 2024-02-27
[PKO-35] Reusable package CEL object conditionals

pbabic-redhat:
### Summary
Add reusable CEL expressions (macros) to package manifests that can be used in CEL annotations.
CEL Macros can be specified here: `PackageManifest.Spec.CelMacros`. Each macro has to have a unique name and the expression has to return a boolean. The expression has access to the full template context. The macros are resolved once per package. When parsing CEL annotations of objects in the package, the macros are replaced with the result of their evaluation (`"true"`, `"false"`).

### Change Type
New Feature

### Check List Before Merging

### Jira
[PKO-35](https://issues.redhat.com/browse/PKO-35)

Documentation is a part of [PKO-62](https://issues.redhat.com/browse/PKO-62)

### Additional Information
This PR depends on [pull/909](https://github.com/package-operator/package-operator/pull/909)


pbabic-redhat:
Please don't get scared of the force-pushing :eyes:. Just made some mistakes when wrestling with the merge conflicts from the cardboard PR.

integration/package-operator/environment.go
```diff
@@ -67,7 +67,7 @@ func init() {
 		panic("PKO_TEST_SUCCESS_MULTI_PACKAGE_IMAGE not set!")
 	}
 	SuccessTestCelPackageImage = os.Getenv("PKO_TEST_SUCCESS_CEL_PACKAGE_IMAGE")
-	if len(SuccessTestMultiPackageImage) == 0 {
+	if len(SuccessTestCelPackageImage) == 0 {
```
erdii, line 70: oopsie :D

erdii:
Hmm.... while I do like the simplicity of the string replacement approach, I don't think it's what we want here. Let me list some pros and cons:

Pros:
- straightforward code
  - first evaluate all the macros
  - then string replace the macros in the cel script text itself
- all macros only get evaluated once

Cons:
- string replacement does not care about exact variable names and happily slices variables into `prefixFOUNDsuffix` where it replaces `FOUND` with either `true` or `false`.
- PKO package code is not so straightforward anymore because the PackageContext is now missing a crucial piece of information to render the objects inside the package: the raw unprocessed macros themselves
- (this is probably not that important performance wise) each macro is executed in its own `cel.Env` (which i suspect is more expensive than injecting + reevaluating the macros for every object)

This diff adds 2 small test cases that showcase how the string replacer trips over code:
```
diff --git a/internal/packages/internal/packagerender/objects_test.go b/internal/packages/internal/packagerender/objects_test.go
index 02943aa3..736d8ba8 100644
--- a/internal/packages/internal/packagerender/objects_test.go
+++ b/internal/packages/internal/packagerender/objects_test.go
@@ -104,6 +104,22 @@ func TestReplaceMacros(t *testing.T) {
 			},
 			result: "(true || false) && false",
 		},
+		{
+			name:       "tricking the replacer - symbol name slicing",
+			expression: "justnotTrue",
+			macros: map[string]string{
+				"notTrue": "false",
+			},
+			result: "justnotTrue",
+		},
+		{
+			name:       "tricking the replacer - dot path slicing / map shadowing",
+			expression: "spec.thing",
+			macros: map[string]string{
+				"spec": "false",
+			},
+			result: "spec.thing",
+		},
 	} {
 		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
```

erdii:
I think we should explore other options for exposing the macros to the cel scripts.

internal/packages/internal/packagerender/celctx/cel.go
```diff
@@ -0,0 +1,142 @@
...
+
+	return out.Value().(bool), nil
+}
+
+func newErrInvalidReturnType(actual, expected *cel.Type) error {
+	return fmt.Errorf("%w: %v, expected %v", errInvalidReturnType, actual, expected)
+}
+
+func unpackContext(tmplCtx *packagetypes.PackageRenderContext) (map[string]any, []cel.EnvOption, error) {
+	// TODO ask why?
```
erdii, line 108: @pbabic-redhat Do you know why and how tmplCtx is supposed to be nil here?

internal/packages/internal/packagerender/celctx/cel.go
```diff
@@ -0,0 +1,142 @@
...
+	if tmplCtx == nil {
+		return map[string]any{}, []cel.EnvOption{}, nil
+	}
```
erdii, line 111: @pbabic-redhat Do you know why and how tmplCtx is `supposed` to be nil here?

pbabic-redhat: ATM, this is only used in tests. However, since the function receives a pointer, I thought it might be safer to check it. That way, it's less susceptible to breaking in the future. But I can totally remove it and modify the tests. WDYT?

pbabic-redhat:
Looking into the coverage now...

internal/packages/internal/packagerender/celctx/cel.go
erdii: Ah understood!

One could argue that swallowing an error state (tmplCtx being a nil-pointer) and covering it up by returning an empty map is surprising and "broken" behavior if we don't expect `tmplCtx` to be ever actually be a nil-pointer.

I'm not so sure if it's worth removing it - use your best judgement. :)

PS: while we're already chatting here - I've talked to @thetechnick about the PR and we would like to request some API changes. We think it makes sense to structure and name the api (combined with the folder conditions PR) like this:

```
spec:
  conditionals:
    snippets:
    - name: xxx
      expression: yyy != zzz
    paths:
    - glob: 'foo/bar/**'
      expression: snippet.xxx && answer == 42 # use the snippet
```

effectively:
- moving the api under an optional key `.spec.conditionals`
- changing the name `macros` to `snippets`
- namespacing the snippet result values into `snippets` in the cel env

an alternative naming convention could look like this:

```
spec:
  conditionalFiltering:
    namedConditions:
    - name: xxx
      expressoin: yyy != zzz
    globs:
    - glob: 'foo/bar/**'
      expression: cond.xxx && answer == 42 # use the named condition
```

Do you think that makes sense? Pls let me know if you want help implementing the changes.

pbabic-redhat: Thanks for the feedback @erdii! Sorry I took so long to respond. I really like the idea of grouping the features into a single field. Together with the renaming in the first sample, I think it makes for a more intuitive UX.

I have a question regarding the snippet namespacing. What is the motivation behind it? I understand that it would force the users to be more explicit, but it would also require them to write more boilerplate. So far, there are 4 template context variables injected into the environment. Do you think these are likely to clash with desirable snippet names?

erdii: Thanks a lot! :) And you do not have to exuse yourself for taking some time. You're great at this and your pacing is too! :)

### Namespacing

Cons:
- More boilerplate code for users because they have to prefix `namespace.` to their condition names.

Pros:
- Backwards compatibility: We can easily add more names without breaking backwards compatiblity with packages that use the same name for a condition.
- Less surprising: A user won't have to question if a variable is an expression they can control (`namespace.foo`) or something that PKO injects for them.

The con can be countered by using a short namespace name. The backwards compatibility pro is a must have.

---

Now that the important stuff has been written - let's do some bikeshedding:

### Naming things

I'd prefer this naming convention
```
spec:
  conditionalFiltering:
    namedConditions:
    - name: xxx
      expression: yyy != zzz
    paths:
    - glob: 'foo/bar/**'
      expression: cond.xxx && answer == 42
```

over the first one because:
- we are using conditional expressions to filter a file tree => "Conditional Filtering" is a very unsurprising name for this
- we are optionally giving conditions a reusable name and store them in a list -> namedCondtions
- the conditions are stored and injected in a map `cond` which is short and concise.

Whereas in the first convention
- the key `conditionals` tells us what we are specifying but not what it does.
- the key `snippets` tells us the right thing (reusable code pieces) but it rubs me the wrong way because this code is only every evaluated once and then stuffed in a variable of the same name, so is it really reusable? kind of yes, kind of no. :shrug:

erdii (approved):

PR #1033, opened 2024-03-19
[PKO-13] Automatic Big Package Support - add BinpackNextFitChunker

erdii:

### Summary

This PR adds a chunker that put's at most 1 Megabyte (1 MiB) of objects into a single ObjectSlice and makes PKO behave as described in [1]:

> When using the 'Package' API, Package Operator will automatically split packages that reach a certain limit by using the 'ObjectSlice' API.

The chunker uses next fit binpacking aka:
- keep 1 bucket open
- put objects into the open bucket
- close the bucket and create a new bucket if the next object is too big

### Change Type

Breaking Change
New Feature

### Check List Before Merging

- [x] Clarify: will this chunker be used by default?

### Additional Information

[1] https://package-operator.run/docs/concepts/big-packages/

### DIFF
--- a/internal/packages/internal/packagedeploy/chunking.go
+++ b/internal/packages/internal/packagedeploy/chunking.go
@@ -2,6 +2,8 @@ package packagedeploy
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 
 	corev1alpha1 "package-operator.run/apis/core/v1alpha1"
 	"package-operator.run/internal/adapters"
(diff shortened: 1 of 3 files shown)


erdii:
The council has decided to:
- Put the `.` in a new line.
- Enable the new chunker by default.

internal/packages/internal/packagedeploy/chunking.go
```diff
@@ -44,8 +64,12 @@ func determineChunkingStrategyForPackage(pkg adapters.GenericPackageAccessor) ob
 	switch chunkingStrategy(strategy) {
 	case chunkingStrategyEachObject:
 		return &EachObjectChunker{}
-	default:
+	case chunkingStrategyBinpackNextFit:
+		return &BinpackNextFitChunker{}
```
kostola, line 68: nit: isn't this redundant given that the `default` branch also returns `&BinpackNextFitChunker{}`?

eqrx (dismissed):

internal/packages/internal/packagedeploy/chunking.go
eqrx: I find it actually good for understanding, the case is explicitly considered and by coincidence does the same as "anything else"

erdii: I'll keep it :)

eqrx (approved):

PR #1045, opened 2024-03-28
[ci/testing] only fail ci on failed code coverage upload when not in PR

erdii:
### Check List Before Merging

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -43,7 +43,7 @@ jobs:
       with:
         files: .cache/unit/cover.txt,.cache/integration/cover.txt
         token: ${{ secrets.CODECOV_TOKEN }}
-        fail_ci_if_error: ${{ github.repository_owner == 'package-operator' }}
+        fail_ci_if_error: ${{ github.event_name != 'pull_request' }}
         verbose: true
 
     - name: Archive unit test results


.github/workflows/testing.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
       with:
         files: .cache/unit/cover.txt,.cache/integration/cover.txt
         token: ${{ secrets.CODECOV_TOKEN }}
-        fail_ci_if_error: ${{ github.repository_owner == 'package-operator' }}
+        fail_ci_if_error: ${{ github.event_name != 'pull_request' }}
```
kostola, line 46: shouldn't it be the opposite?
```yaml
fail_ci_if_error: ${{ github.event_name == 'pull_request' }}
```

erdii: No. CI should NOT fail when this workflow is run during a pull request and this action error.

thus we want:
- for PR -> `fail_ci_if_error: false`
- for main branch tests and other things -> `fail_ci_if_error: true`

kostola: my bad...I misread the title of the PR and thought it was the opposite

kostola (approved):
It should work...let's see 😜

PR #1111, opened 2024-04-19
PKO-82: Implement regression test and bugfix for invalid remote-phase package upgrade

erdii:
In a previous release, the remote-phase component was moved from the top-level component of its own package to the `remote-phase` component of the package-operator package.

Upgrading all existing remote-phase Packages controlled by PKO includes the (up until now missing pieces):
- diffing existing-package vs desired-package `.spec` and not only `.spec.image`
- updating/overriding `.spec` (and thus also, the crucial field: `.spec.component`) instead of only updating `.spec.image`

### Change Type

Bug Fix

### Check List Before Merging


kostola (approved):
LGTM

PR #1137, opened 2024-05-06
[PKO-41] Print affirmative message when package validation succeeds

Ankit152:
### Summary

Resolves: [PKO-41](https://issues.redhat.com/browse/PKO-41)

### Check List Before Merging


erdii (changes requested):
Hey Ankit :wave:
Thank you very much for the contribution! :)

Would you mind adding a regression test for this to our integration test suite please?

cmd/kubectl-package/validatecmd/validate.go
```diff
@@ -53,6 +54,8 @@ func NewCmd(validator Validator) *cobra.Command {
 			return fmt.Errorf("validating package: %w", err)
 		}
 
+		log.Println("Package validated successfully!")
```
eqrx, line 57: Please use OutOfStdout for producing output! https://pkg.go.dev/github.com/spf13/cobra#Command.OutOrStdout

eqrx (dismissed):

cmd/kubectl-package/validatecmd/validate.go
Ankit152: Okay

eqrx: Sorry, forgot to write down why :D

Reason is that `log.Println` simply prints to stdout. When a cobra command is invoked it is possible to specify `io.Writer` for stdout and stderr which makes testing much easier!

erdii: TIL!

erdii:
Yes that looks like the right file! I'll take a look at what you added to this PR right away :)

cmd/kubectl-package/validatecmd/validate.go
```diff
@@ -53,6 +53,8 @@ func NewCmd(validator Validator) *cobra.Command {
 			return fmt.Errorf("validating package: %w", err)
 		}
 
+		fmt.Fprintln(cmd.OutOrStdout(), "Package validated successfully!")
```
erdii, line 56: Do you think offloading this into a `const` at the top of the file would make sense?
My thought is: I'ts not a magic string per-say but having a name like `validationSuccessMessage` highlights its semantical meaning. And also highlights that this string will become part of the cli "API".

WDYT?

erdii (changes requested):
The changed int tests look good. Although the'res a unit test that needs adaption too:

```
📦 package-operator.run/cmd/kubectl-package/validatecmd (24.2% coverage)
  ❌ TestValidateFolder (20ms)
    validate_test.go:26: 
        	Error Trace:	/home/erdii/projects/redhat/github.com/package-operator/package-operator/cmd/kubectl-package/validatecmd/validate_test.go:26
        	Error:      	Should be empty, but was Package validated successfully!
        	Test:       	TestValidateFolder
```

cmd/kubectl-package/validatecmd/validate.go
Ankit152: Yup, it makes sense to me as well.

erdii: image

erdii: looks good to me! :) Thanks

erdii (approved):

PR #1229, opened 2024-06-11
feat: adding missing unit tests for pko commands

Ankit152:
### Check List Before Merging

### DIFF
--- a/cmd/kubectl-package/updatecmd/update_test.go
+++ b/cmd/kubectl-package/updatecmd/update_test.go
@@ -1,0 +1,80 @@
+package updatecmd
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/require"
+
+	internalcmd "package-operator.run/internal/cmd"
+)
+
+func TestUpdate_WithoutPath(t *testing.T) {
+	t.Parallel()
+
+	cmd := NewCmd(internalcmd.NewUpdate(internalcmd.WithLog{
+		Log: logr.Logger{},
+	}))
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
+	cmd.SetOut(stdout)
+	cmd.SetErr(stderr)
+
+	require.Error(t, cmd.Execute())
+	require.NotEmpty(t, stderr.String())
+}
+
+func TestUpdate_WithInvalidPath(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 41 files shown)


Ankit152:
@thetechnick can you have a look at this PR and this one too #1258 ?

cmd/kubectl-package/updatecmd/update_test.go
```diff
@@ -0,0 +1,42 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/require"
+
+	internalcmd "package-operator.run/internal/cmd"
+)
+
+func TestUpdate_WithoutPath(t *testing.T) {
```
kostola, line 13: I think we should also add tests for the happy paths:
1. Package gets successfully updated
2. Package is already up-to-date

Ankit152: I have added the the test for Package is already up-to-date.
But if I do a test for package getting successfully updated then every time it will perform a write operation which will not pass CI Jobs.

kostola: You could have another simple package pointing to a well known package image. This image would have an old (fake?) hash in the lock file and running the test would update it.
I don't understand how this would make the CI jobs fail.

Ankit152: Hey @kostola! 
Right now, when we try to update a package we don't see any affirmative message when the package gets successfully updated. IMO, we should create a small ticket for adding an affirmative message for successful update of a package along with its unit test. WDYT?

kostola: > IMO, we should create a small ticket for adding an affirmative message for successful update of a package along with its unit test. WDYT?

@Ankit152 fine for me but I won't consider it at all for this PR. IMHO you can implement the test I suggested in [the comment above](https://github.com/package-operator/package-operator/pull/1229#discussion_r1713263458) without this change.

Ankit152: I have added the tests.

Ankit152: I have found a work-around for this though. Let me know, if you have any thoughts.

Ankit152: @kostola there was an issue in golang-ci due to which the tests were failing. Can you please have a look at the tests now?

kostola: it looks good now :+1:

PR #1258, opened 2024-06-26
Added missing unit tests for manager and adapters

Ankit152:
### Check List Before Merging

### DIFF
--- a/cmd/package-operator-manager/bootstrap/apihelper_test.go
+++ b/cmd/package-operator-manager/bootstrap/apihelper_test.go
@@ -86,4 +86,146 @@ func TestIsPKOAvailable(t *testing.T) {
 		require.NoError(t, err)
 		assert.True(t, isPKOAvailable)
 	})
+
+	t.Run("not available - available replicas is 0", func(t *testing.T) {
+		t.Parallel()
+		c := testutil.NewClient()
+
+		c.On("Get", mock.Anything, mock.Anything,
+			mock.AnythingOfType("*v1.Deployment"),
+			mock.Anything).
+			Run(func(args mock.Arguments) {
+				depl := args.Get(2).(*appsv1.Deployment)
+				depl.Status.AvailableReplicas = 0
+			}).
+			Return(nil)
+
+		isPKOAvailable, err := isPKOAvailable(
+			context.Background(), c, "")
+		require.NoError(t, err)
+		assert.False(t, isPKOAvailable)
+	})
+
+	t.Run("not available - available replicas is not equal to available replicas", func(t *testing.T) {
+		t.Parallel()
+		c := testutil.NewClient()
+
+		c.On("Get", mock.Anything, mock.Anything,
+			mock.AnythingOfType("*v1.Deployment"),
+			mock.Anything).
+			Run(func(args mock.Arguments) {
+				depl := args.Get(2).(*appsv1.Deployment)
+				depl.Status.AvailableReplicas = 2
+				depl.Status.UpdatedReplicas = 1
+			}).
+			Return(nil)
+
+		isPKOAvailable, err := isPKOAvailable(
+			context.Background(), c, "")
+		require.NoError(t, err)
+		assert.False(t, isPKOAvailable)
+	})
+}
+
+func TestIsPKOClusterPackageAvailable(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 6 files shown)


cmd/package-operator-manager/bootstrap/apihelper_test.go
```diff
@@ -87,3 +87,201 @@ func TestIsPKOAvailable(t *testing.T) {
...
+				depl.Status.UpdatedReplicas = depl.Status.AvailableReplicas
+			}).
+			Return(nil)
+
+		isPKODeploymentAvailable, err := isPKODeploymentAvailable(
+			context.Background(), c, "")
+		require.NoError(t, err)
+		assert.True(t, isPKODeploymentAvailable)
+	})
+}
```
kostola, line 287: I might be wrong but this to me looks identical to https://github.com/package-operator/package-operator/blob/7186448b9e86a9920496b3d3155c8e8e117e5669/cmd/package-operator-manager/bootstrap/apihelper_test.go#L20-L89

cmd/package-operator-manager/components/components_test.go
```diff
@@ -34,3 +34,24 @@ func TestUncachedClient(t *testing.T) {
 	require.EqualError(t, err,
 		"unable to set up uncached client: must provide non-nil rest.Config to client.New")
 }
+
+func TestProvideRestConfig(t *testing.T) {
+	t.Parallel()
+
+	_, err := ProvideRestConfig()
+	require.Error(t, err)
```
kostola, line 42: What error are we expecting here?

cmd/package-operator-manager/components/components_test.go
```diff
@@ -34,3 +34,24 @@ func TestUncachedClient(t *testing.T) {
...
+
+	_, err := ProvideRESTMapper()
+	require.NoError(t, err)
+}
+
+func TestProvideManager(t *testing.T) {
+	t.Parallel()
+
+	_, err := ProvideManager(nil, nil, Options{})
+	require.Error(t, err)
```
kostola, line 56: What error are we expecting here?

cmd/package-operator-manager/components/options_test.go
```diff
@@ -67,3 +67,15 @@ func TestProvideOptions(t *testing.T) {
 		},
 	}, opts)
 }
+
+func TestEnvToInt(t *testing.T) {
```
kostola, line 71: I think there should be two more cases
1. an existing env variable successfully converted to int
2. an existing env variable containing a random string that triggers parse error

cmd/package-operator-manager/bootstrap/apihelper_test.go
Ankit152: Here as the name suggests, we are testing for `isPKODeploymentAvailable` which is missing. The snippet you are referring is for `isPKOAvailable`. 🙂

cmd/package-operator-manager/components/components_test.go
Ankit152: These errors are left over the way other tests in file has been left.

cmd/package-operator-manager/components/components_test.go
Ankit152: same as above.

cmd/package-operator-manager/components/options_test.go
Ankit152: Added the test cases.

cmd/package-operator-manager/bootstrap/apihelper_test.go
kostola: `isPKOAvailable` is internally calling both `isPKOClusterPackageAvailable` and `isPKODeploymentAvailable`.

I think these tests could be rewritten by just calling `isPKOAvailable` including all the possible test cases **only once** (the first two test cases in `TestIsPKOAvailable` and `TestIsPKODeploymentAvailable` are _exactly_ the same).

The coverage would be the same but the readability would drastically improve.

Ankit152: Done.

kostola: ok looks better now

kostola (approved):
lgtm 👍

PR #1286, opened 2024-07-15
[PKO-106] fix the endless reconcilation

venkateshsredhat:
https://issues.redhat.com/browse/PKO-106
### Summary

Currently the object template controller updates the object template object on every reconcile regardless of whether the desired state is equal to the current state of the object in the cluster. This creates unnecessary load on the api-server.[Ref: [https://github.com/package-operator/package-operator/blob/main/internal/controllers/objecttemplate/template_reconciler.go#L109]](https://issues.redhat.com/browse/PKO-106#L109%5D)

How this PR will help :

Currently in the code we are trying to make the update call until the spec gets changed . Requesting your review since the it seems to be a Unstructured Object so taking the "spec" value from the Map .
### Check List Before Merging

### DIFF
--- a/internal/controllers/objecttemplate/template_reconciler.go
+++ b/internal/controllers/objecttemplate/template_reconciler.go
@@ -102,24 +102,32 @@ func (r *templateReconciler) Reconcile(
 	if err := updateStatusConditionsFromOwnedObject(ctx, objectTemplate, existingObj); err != nil {
 		return res, fmt.Errorf("updating status conditions from owned object: %w", err)
 	}
-	ownedByPKO := utils.HasSameController(existingObj, obj)
-	for field, currentValue := range obj.Object {
+
+	if handleUpdate(existingObj, obj) {
+		obj.SetOwnerReferences(existingObj.GetOwnerReferences())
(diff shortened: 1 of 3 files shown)


venkateshsredhat:
addressing comments from the PR https://github.com/package-operator/package-operator/pull/1255 .

internal/controllers/objecttemplate/template_reconciler.go
```diff
@@ -100,14 +102,21 @@ func (r *templateReconciler) Reconcile(
...
-
-	obj.SetOwnerReferences(existingObj.GetOwnerReferences())
-	obj.SetLabels(labels.Merge(existingObj.GetLabels(), obj.GetLabels()))
-	obj.SetAnnotations(labels.Merge(existingObj.GetAnnotations(), obj.GetAnnotations()))
-
-	obj.SetResourceVersion(existingObj.GetResourceVersion())
-	if err := r.client.Update(ctx, obj); err != nil {
-		return res, fmt.Errorf("updating templated object: %w", err)
+	ownedByPKO := utils.HasSameController(existingObj, obj)
+	for field, currentValue := range obj.Object {
```
ashishmax31, line 106: This can be moved to a separate function

PR #1307, opened 2024-07-25
Kickstart parameterization

thetechnick:
### Check List Before Merging


eqrx (changes requested):

internal/packages/internal/packagekickstart/kickstart.go
```diff
@@ -22,22 +27,38 @@ type KickstartResult struct {
 	GroupKindsWithoutProbes []schema.GroupKind
 }
 
-func Kickstart(_ context.Context, pkgName string, objects []unstructured.Unstructured) (
+func Kickstart(
+	_ context.Context, pkgName string,
+	objects []unstructured.Unstructured,
+	paramFlags []string,
```
kostola, line 33: Nit: I find a bit unclear the fact that this param, which is a slice of strings, is called `paramOpts` before this invocation all the way up but in this function it becomes `paramFlags` and `paramOpts` becomes the actual parsed object.

I would stick with `paramOpts` for the slice of strings and use a different name for the parsed objects (like `paramOptsObj` or `parsedOpts`), just for consistency.

internal/packages/internal/packagekickstart/presets/param.go
```diff
@@ -0,0 +1,96 @@
...
+	tolerationsParam   = "tolerations"
+	nodeSelectorsParam = "nodeselectors"
+	resourcesParam     = "resources"
+	envParam           = "env"
+	imagesParam        = "images"
+	allParam           = "all"
+)
+
+func ParametrizeOptionsFromFlags(opts []string) ParametrizeOptions {
+	if slices.Contains(opts, "all") {
```
kostola, line 38: Should we throw an error if `all` is passed together with any other value? It may hint an error in the command call. Wdyt?

PR #1314, opened 2024-08-02
[PKO-106] fix the endless reconcilation #1286

venkateshsredhat:
https://issues.redhat.com/browse/PKO-106

Summary
Currently the object template controller updates the object template object on every reconcile regardless of whether the desired state is equal to the current state of the object in the cluster. This creates unnecessary load on the api-server.[Ref: [https://github.com/package-operator/package-operator/blob/main/internal/controllers/objecttemplate/template_reconciler.go#L109]](https://issues.redhat.com/browse/PKO-106#L109%5D)

How this PR will help :

Currently in the code we are trying to make the update call until the spec gets changed . Requesting your review since the it seems to be a Unstructured Object so taking the "spec" value from the Map .

### Check List Before Merging

### DIFF
--- a/internal/controllers/objecttemplate/template_reconciler.go
+++ b/internal/controllers/objecttemplate/template_reconciler.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/util/jsonpath"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -31,6 +32,8 @@ import (
 
 // Requeue every 30s to check if input sources exist now.
 var defaultMissingResourceRetryInterval = 30 * time.Second
+
+const FieldOwner = "package-operator"
 
 type templateReconciler struct {
 	*environment.Sink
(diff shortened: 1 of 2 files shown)


internal/controllers/objecttemplate/template_reconciler.go
```diff
@@ -106,6 +111,25 @@ func (r *templateReconciler) Reconcile(
...
+		types.ApplyPatchType, objectPatch),
+		client.FieldOwner(FieldOwner),
+		client.ForceOwnership,
+	); err != nil {
+		return res, fmt.Errorf("patching object: %w", err)
+	}
+
 	if err := r.client.Update(ctx, obj); err != nil {
 		return res, fmt.Errorf("updating templated object: %w", err)
 	}
```
venkateshsredhat, line 135: i will remove this

internal/controllers/objecttemplate/template_reconciler.go
```diff
@@ -106,10 +112,26 @@ func (r *templateReconciler) Reconcile(
 	obj.SetAnnotations(labels.Merge(existingObj.GetAnnotations(), obj.GetAnnotations()))
 
 	obj.SetResourceVersion(existingObj.GetResourceVersion())
-	if err := r.client.Update(ctx, obj); err != nil {
-		return res, fmt.Errorf("updating templated object: %w", err)
+	if err := r.fixFieldManagers(ctx, existingObj); err != nil {
```
ashishmax31, line 115: On second thought this might not be required since PKO creates the object. So the field manager field is not set by any client side apply.

ashishmax31: Wdyt @erdii

ashishmax31, line 115: On second thought this might not be required since PKO creates/updates these objects using not create and update calls. No CSA apply fields would be present. wdyt @erdii

internal/controllers/objecttemplate/template_reconciler.go
```diff
@@ -106,10 +112,26 @@ func (r *templateReconciler) Reconcile(
...
 
+	patch := obj.DeepCopy()
+	unstructured.RemoveNestedField(patch.Object, "status")
+
+	objectPatch, err := json.Marshal(patch)
+	if err != nil {
+		return res, fmt.Errorf("creating patch: %w", err)
+	}
+	log.Println("Entering patch:")
+	if err := r.client.Patch(ctx, existingObj, client.RawPatch(
```
ashishmax31, line 127:
```suggestion
	if err := r.client.Patch(ctx, obj, client.Apply, &client.PatchOptions{
		FieldManager: FieldOwner,
		Force:        ptr.To(true),
	}); err != nil {
	....
	}
```

PR #1360, opened 2024-08-14
Refactor common constants into a "constants" package and make preflight/dryrun use constant.FieldOwner

erdii:

### Change Type

Refactored common constants into their own package with no other imports so that preflight/dryrun.go can import `constants.FieldOwner` without running into a circular dependency.

Refactoring/Cleanup

### Check List Before Merging


kostola:
I like the idea!

@erdii what are the guidelines to put a constant in the `constants` package? If it's used in more than one sub-package?

kostola (approved):
My [previous comment](https://github.com/package-operator/package-operator/pull/1360#issuecomment-2288312672) was more related to future development, which we can discuss later.

Overall this PR looks good to me, so it's worth an approve 😜

erdii:
Thanks for your review! :heart: 

> @erdii what are the guidelines to put a constant in the constants package? If it's used in more than one sub-package?

Yeah. I just moved all the consts from the controllers package, most of which get used all over the pko codebase into a separate package, so the rule of thumb is: If you introduce a const, place it at the closest place possible that is still visible to all the places that want to use it.

PR #1382, opened 2024-08-27
[PKO-60] Do not attempt to overwrite full object on teardown of object which we don't control anymore

weherdh:
### Summary

Update the way to patch the object by removing `ownerReference` of the teardown external object

This resolves the https://issues.redhat.com/browse/PKO-60 
### Change Type

Bug fix

### Check List Before Merging

### Additional Information

```
=== RUN   TestPackage_externalObject_teardown
    wait.go:153: "level"=0 "msg"="waiting 40s on package-operator.run/v1alpha1, Kind=Package default/success to report condition \"Unpacked\"=\"True\"..."
    wait.go:153: "level"=0 "msg"="waiting 40s on package-operator.run/v1alpha1, Kind=Package default/success to report condition \"my-prefix/Progressing\"=\"True\"..."
    wait.go:153: "level"=0 "msg"="waiting 40s on package-operator.run/v1alpha1, Kind=Package default/success to report condition \"Available\"=\"True\"..."
    wait.go:153: "level"=0 "msg"="waiting 20s on package-operator.run/v1alpha1, Kind=ObjectSet default/external to report condition \"Available\"=\"True\"..."
    wait.go:153: "level"=0 "msg"="waiting 15s on apps/v1, Kind=Deployment default/test-stub-success external ownerReference to be removed..."
    wait.go:193: "level"=0 "msg"="waiting 20s for package-operator.run/v1alpha1, Kind=ObjectSet default/external to be gone..."
    wait.go:193: "level"=0 "msg"="waiting 20s for package-operator.run/v1alpha1, Kind=Package default/success to be gone..."
--- PASS: TestPackage_externalObject_teardown (17.28s)
PASS
```

### DIFF
--- a/integration/package-operator/package_test.go
+++ b/integration/package-operator/package_test.go
@@ -7,6 +7,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"testing"
+	"time"
+
+	"k8s.io/utils/ptr"
 
 	"github.com/go-logr/logr"
 	"github.com/go-logr/logr/testr"
@@ -17,7 +20,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"pkg.package-operator.run/cardboard/kubeutils/wait"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"package-operator.run/internal/constants"
 
 	corev1alpha1 "package-operator.run/apis/core/v1alpha1"
 	manifestsv1alpha1 "package-operator.run/apis/manifests/v1alpha1"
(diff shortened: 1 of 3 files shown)


integration/package-operator/package_test.go
```diff
@@ -302,3 +303,117 @@ func TestPackage_nonExistent(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, "ImagePullBackOff", existingPackage.Status.Conditions[0].Reason)
 }
+
+func TestPackage_externalObject_teardown(t *testing.T) {
```
Ankit152, line 307: The name of the test should be `object_teardown` instead of `externalObject` as you are making changes in `objects`.
External Object is something I am working on. 🙂

weherdh: Am I wrong?

weherdh: Whoops, I think I mixed up yours and mine :expressionless:. What we can do now :joy:

Ankit152: Just rename the test.

weherdh: I added the internal test below

internal/controllers/phase_reconciler.go
```diff
@@ -361,14 +361,36 @@ func (r *PhaseReconciler) teardownPhaseObject(
...
+		r.ownerStrategy.RemoveOwner(owner.ClientObject(), object)
+		objectPatch, err := json.Marshal(object)
+		if err != nil {
+			return false, fmt.Errorf("creating patch: %w", err)
+		}
+		if err = r.writer.Patch(ctx, currentObj, client.RawPatch(
+			types.MergePatchType, objectPatch,
+		)); err != nil {
+			return false, fmt.Errorf("removing external object owner reference: %w", err)
+		}
```
erdii, line 376: This is not an SSA (as requested) for technical reasons, @weherdh outlined in the backing jira issue:

IIUC It is not possible to only remove some known fields from an object using SSA if the full object was created with the same `fieldowner`, as long as the full shape of the object is unknown because every field that is left unspecified in the request, will get removed from the object.

@thetechnick: Would you agree, that using a MergePatch instead is the correct way to go here ?

erdii (commented):
The external object teardown change is not needed anymore, since the feature was removed in #1442

internal/controllers/phase_reconciler.go
thetechnick: Merge patch is fine for this case, in the end we just want to surgically remove the owner and cache label and nothing else.
But the label is not yet removed here, just the owner.

ashishmax31:
@weherdh You need to sort the imports in the file integration/package-operator/package_test.go. That should fix the CI issue.

erdii:
I took the liberty to clean up this PR and rebase it against main.

In case we need the old work back - last commit by Wenqi: `dce244e596cbbeea15d207361d11aa1a9d0cf8a0`

erdii:
Nevermind. I don't have the permissions to update this PR. Opening a new one.

PR #1448, opened 2024-10-04
ref: remove ctx as it's not being used

boranx:
### Summary
playing w/ the code-base

### Change Type
Refactor

Do you know if we need the context?
(I see we omit when passing the object at first glance)

### Check List Before Merging


eqrx (approved):
nice, thanks :)

erdii (approved):

PR #1504, opened 2024-11-05
chore: does anyone know why don't we call walkWithSymlinks directly?

boranx:


erdii:
IIUC then `walk` is a convenience/secure-by-default wrapper for `walkWithSymlinks` as it ensures that you pass the same dir name as `filename` and `linkDirname` arg which effectively constrains the walker to only resolve symlinks with the walked directory tree.

A comment on the func's signature could have helped our understanding though :sweat_smile:

PR #1535, opened 2024-11-13
[PKO-164] Pause Feature for the (Cluster)ObjectDeployment API

pbabic-redhat:
### What
Enable pausing ObjectDeployment reconciliation.

### Why
As an SRE working with package operator, I want an easy way to pause the reconciliation of a (Cluster)ObjectDeployment to perform some debugging or do in-place edits that would otherwise be reset.

### How
* Add a new field .spec.paused to (Cluster)ObjectDeployment APIs.
* Update ObjectDeployment controllers to propagate pause down to the (Cluster)ObjectSet.
* The ObjectDeployment's paused condition MUST properly reflect the reported pause status of all api levels underneath
* The pause value in the ObjectDeployment controls the pause value in ObjectSet. If ObjectSet is manually changed, the value will be reverted to the one in the owner ObjectDeployment.
* (Cluster)ObjectDeployments will not update or rollout (Cluster)ObjectSets when paused.

[Demo recording](https://drive.google.com/file/d/18ivurVkkGRPZ16uyUKcpqNfv1d_Kftld/view?usp=drive_link)

### Change Type
* New Feature

### Jira
[PKO-164](https://issues.redhat.com/browse/PKO-164)

### Check List Before Merging
- [x] This PR passes all pre-commit hook validations.
- [x] This PR is fully tested and regression tests are included.
- [x] Relevant documentation has been updated.

PR #1567, opened 2024-11-27
[PKO-136] Pause feature for (Cluster)Package API

radimous:
### Change Type
New Feature

### Jira
[PKO-136](https://issues.redhat.com/browse/PKO-136)

### Check List Before Merging

      standard.

### Additional Information

Depends on #1535


thetechnick (dismissed):
Good work! :+1:

internal/adapters/package.go
```diff
@@ -176,7 +186,20 @@ func (a *GenericClusterPackage) GetUnpackedHash() string {
...
+func (a *GenericClusterPackage) GetSpecPaused() bool {
+	return a.Spec.Paused
+}
+
+func (a *GenericClusterPackage) SetSpecPaused(paused bool) {
+	a.Spec.Paused = paused
+}
+
 func updatePackagePhase(pkg GenericPackageAccessor) {
+	if meta.IsStatusConditionTrue(*pkg.GetConditions(), string(corev1alpha1.PackagePaused)) {
```
pbabic-redhat, line 198: I think this is not needed.
```suggestion
	if meta.IsStatusConditionTrue(*pkg.GetConditions(), corev1alpha1.PackagePaused) {
```

internal/controllers/packages/package_controller_test.go
```diff
@@ -0,0 +1,262 @@
+package packages
```
pbabic-redhat, line 1: The unit tests here look really good, thanks! But there is no test for the pause feature. I think that would be nice to have.

pbabic-redhat (changes requested):
Great work!

I left a couple of comments. Please let me know if I can help with anything. :)

config/packages/test-stub-pause/manifest.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: manifests.package-operator.run/v1alpha1
```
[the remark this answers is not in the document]
pbabic-redhat: This is also a detail. Please consider it more of a feedback rather than a must have change. :grin:

internal/controllers/packages/object_deployment_status_reconciler.go
```diff
@@ -43,6 +44,16 @@ func (r *objectDeploymentStatusReconciler) Reconcile(
...
 	}
 
+	objDepPausedCond := meta.FindStatusCondition(*objDep.GetConditions(), corev1alpha1.ObjectDeploymentPaused)
+	if objDepPausedCond != nil && objDepPausedCond.ObservedGeneration == objDep.ClientObject().GetGeneration() &&
+		objDepPausedCond.Status == metav1.ConditionTrue {
+		packagePausedCond := objDepPausedCond.DeepCopy()
+		packagePausedCond.ObservedGeneration = packageObj.ClientObject().GetGeneration()
+		meta.SetStatusCondition(packageObj.GetConditions(), *packagePausedCond)
+	} else {
+		meta.RemoveStatusCondition(packageObj.GetConditions(), corev1alpha1.PackagePaused)
```
erdii, line 54: I don't think we should remove the status condition here. We should keep it and explicitly set it to false.

erdii (changes requested):
lgtm otherwise!

internal/controllers/packages/object_deployment_status_reconciler.go
erdii: ah whatever - i don't need to be nitpicky

erdii (approved):

PR #1662, opened 2025-01-02
Update to go 1.23

eqrx:


eqrx:
(https://github.com/kubernetes-sigs/controller-tools/pull/877 removed part of the controller-gen generated docs)

eqrx:
Comments removed with the version bump all seem to be coming from outside of pko, so the drop is desired

.golangci.yaml
```diff
@@ -152,6 +152,7 @@ linters-settings:
     local-prefixes: package-operator.run
   gosec:
     excludes:
+    - G115 # TODO: Lossy integer conversions.
```
pbabic-redhat, line 155: Leftover TODO?

cmd/package-operator-manager/bootstrap/bootstrap_test.go
```diff
@@ -96,10 +96,7 @@ func TestBootstrapper_bootstrap(t *testing.T) {
 	c := testutil.NewClient()
 	b := &Bootstrapper{client: c}
 
-	var (
-		runManagerCalled bool
-		runManagerCtx    context.Context
-	)
```
pbabic-redhat, line 102: Why separate these? They're used in the same place.

config/images/app-interface-push-images.Containerfile
```diff
@@ -0,0 +1,20 @@
+FROM registry.access.redhat.com/ubi9/ubi:latest
```
pbabic-redhat, line 1: Isn't this file supposed to be in downstream? Also, is it tested somehow, where is it used?

pbabic-redhat (commented):
Hi @eqrx!
Left a couple of comments.

cmd/package-operator-manager/bootstrap/bootstrap_test.go
```diff
@@ -127,6 +124,7 @@ func TestBootstrapper_bootstrap(t *testing.T) {
 	ctx, cancel := context.WithTimeout(
 		context.Background(), testBootstrapTimeout)
 	defer cancel()
+	var runManagerCtx context.Context
```
erdii, line 127: If this is moved down here, then I'd argue that `runManagerCalled` should be too

config/images/app-interface-push-images.Containerfile
erdii: Yup - this file doesn't belong here

erdii (changes requested):
Should this PR bump the `toolchain` (and `go`?) directives in our `go.mod` and `go.work` files?

eqrx:
> Should this PR bump the `toolchain` (and `go`?) directives in our `go.mod` and `go.work` files?

Unsure what you mean, yes it is supposed to set go1.23 as minimum required so it needs to

erdii:
> > Should this PR bump the `toolchain` (and `go`?) directives in our `go.mod` and `go.work` files?
> 
> Unsure what you mean, yes it is supposed to set go1.23 as minimum required so it needs to

At the time of writing no go.mod/work files had been touched by this PR, so I suggested it.

apis/go.mod
```diff
@@ -1,8 +1,6 @@
 module package-operator.run/apis
 
-go 1.22.0
-
-toolchain go1.22.2
```
erdii, line 5: Why is the toolchain directive removed?

go.mod
```diff
@@ -1,8 +1,6 @@
 module package-operator.run
 
-go 1.22.0
-
-toolchain go1.22.2
```
erdii, line 5: Same question re toolchain

pkg/go.mod
```diff
@@ -1,8 +1,6 @@
 module package-operator.run/pkg
 
-go 1.22.0
-
-toolchain go1.22.2
```
erdii, line 5: Same question re toolchain

go.work
```diff
@@ -1,6 +1,6 @@
-go 1.22.0
+go 1.23
 
-toolchain go1.22.2
```
erdii, line 3: And here toolchain is kept?

erdii (commented):
Why are the toolchain directives inconsistent?
Was this decided by go tooling?

eqrx:
> > > Should this PR bump the `toolchain` (and `go`?) directives in our `go.mod` and `go.work` files?
> > 
> > 
> > Unsure what you mean, yes it is supposed to set go1.23 as minimum required so it needs to
> 
> At the time of writing no go.mod/work files had been touched by this PR, so I suggested it.
> 
> Edit: look - first revision of this PR: [978053d](https://github.com/package-operator/package-operator/commit/978053de0dd9b05b5187aa45eb2745eee6747622)

Ah I see, thx, my bad!

erdii (approved):

PR #1663, opened 2025-01-03
Use golangci-lint v1.63.3

eqrx:


.golangci.yaml
```diff
@@ -155,6 +155,7 @@ linters-settings:
     local-prefixes: package-operator.run
   gosec:
     excludes:
+    - G115 # TODO: Truncating integer conversions.
```
erdii, line 158: What does this TODO mean?

cmd/package-operator-manager/bootstrap/bootstrap_test.go
```diff
@@ -90,6 +90,7 @@ func TestBootstrapperBootstrap(t *testing.T) {
 	assert.Equal(t, "noxxx", os.Getenv("NO_PROXY"))
 }
 
+//nolint:fatcontext
```
erdii, line 93: I'd suggest moving this directly to the offending line and/or explaining why the linter is excluded for the function.

erdii (changes requested):

.golangci.yaml
eqrx: It is an important lint but it is not within the scope of this PR to write helper methods for strict integer conversions so I will leave the check disabled until we come around to that.

cmd/package-operator-manager/bootstrap/bootstrap_test.go
eqrx: Applied

erdii (approved):

PR #1715, opened 2025-01-23
Fix tests that use testify same assertions wrong

eqrx:
For testify to be able to check if two object are the same or not it needs pointers as parameters. Testify skipped the check so far if it got non pointers but stopped doing so: https://github.com/stretchr/testify/issues/1661


internal/packages/internal/packageimport/request_manager_test.go
```diff
@@ -114,7 +114,15 @@ func TestRequestManager_DelayedRequests(t *testing.T) {
 		for j := maxRequestIndex; j >= 0; j-- {
 			af := pkg[i].Files
 			bf := pkg[j].Files
-			assert.NotSame(t, af, bf)
+			// Iterate over all file keys in the first file.
```
erdii, line 117: WDYT about somethink like

```suggestion
			// Iterate over all file keys in the first file map and
			// assert that the backing byte slices in af[key] and bf[key] reference differing slices.
```

and dropping the other comments in the loop?

erdii (approved):

PR #1728, opened 2025-01-29
Enforce go1.23.5

eqrx:
go < 1.23.5 contains vulnerabilites


eqrx:
Superseeded, already in main

PR #1768, opened 2025-02-11
[PKO-208]Add GH Action that runs go mod tidy on dependabot PRs

weherdh:
### Summary

This is to resolve https://issues.redhat.com/browse/PKO-208 for adding a github action to run go mod tidy to keep the module clean

### Change Type

 Bug Fix 

### Check List Before Merging

      standard.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -12,7 +12,65 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  update-go-modules:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '>=1.22'
+          cache-dependency-path: "**/*.sum"
+          check-latest: true
+
+      - name: Detect Go modules
+        id: detect_modules
+        run: |
+          echo "Finding all Go modules..."
+          MODULES=$(find . -mindepth 1 -name "go.mod" | sort | xargs -n1 dirname)
+          MODULES=". $MODULES"
+          MODULES=$(echo "$MODULES" | tr '\n' ' ')  # Convert newlines to spaces
+          echo "Detected modules: $MODULES"
+          echo "MODULES=$MODULES" >> $GITHUB_ENV
+
+      - name: Run go mod tidy for all modules
+        run: |
+          UPDATE=""
+          for module in $MODULES; do
+            echo "Processing module: $module"
+            cd "$module" || exit 1
+            go mod tidy
+            if ! git diff --quiet; then
+              echo "Detected changes in $module"
+              UPDATE="$UPDATE $module"
+              git add go.mod go.sum
+            fi
+            cd - > /dev/null
+          done
+          echo "UPDATE=$UPDATE" >> $GITHUB_ENV
+
+      - name: Run go work sync
+        if: env.UPDATE != ''
+        run: go work sync
+
+      - name: Commit and push changes
+        if: env.UPDATE != ''
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git commit -m "fix: sync dependencies with go mod tidy" || echo "No changes to commit"
+          git push origin HEAD:${{ github.head_ref }}
+
   lint-unit-int:
+    needs: [ update-go-modules ]
+    if: always()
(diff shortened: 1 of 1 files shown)


erdii (changes requested):
Nice approach! I'd like to request a change though:

IIUC then this change skips the `lint-unit-int` job for all, but dependabot PRs. We still need the tests for all other PRs though.

---

I have another idea about reducing code in the pipeline although I'm not quite sure about it, because even though the approach you took duplicates code, it also is very straightforward. So feel free to not implement.

The logic for the root module and for the sub-modules is different. Do you think it would make sense to throw all of them together into a single loop that iterates over `.` + detected submodules? (Btw I really like the approach of dynamically discovering go mods!)
  - Refactoring idea: shared update code for all modules + a single combined git commit
    - set the `UPDATE` to `empty string` prior to the loop
    - run `go mod tidy` for each module
    - then git diff and if the repo is dirty:
      - append the module name to `UPDATE`
      - git stage the current module's go.* files
        - so they will be included in the commit
        - so they will be excluded from the diffs for the other modules
    - Then a single `git commit` in an extra step would suffice.

weherdh:
Thanks @erdii, I am working on addressing your comments

erdii:
Hey @weherdh :  @eqrx just found out that, after running go mod tidy for all modules (Root first + Sub afterwards), we also have to run `go work sync` on the top level module one last time.

.github/workflows/testing.yaml
```diff
@@ -12,7 +12,65 @@ concurrency:
...
+          for module in $MODULES; do
+            echo "Processing module: $module"
+            cd "$module" || exit 1
+            go mod tidy
+            if ! git diff --quiet; then
+              echo "Detected changes in $module"
+              UPDATE="$UPDATE $module"
+              git add go.mod go.sum
+            fi
+            cd - > /dev/null
```
erdii, line 55:
```suggestion
            pushd "$module" || exit 1
            go mod tidy
            if ! git diff --quiet; then
              echo "Detected changes in $module"
              UPDATE="$UPDATE $module"
              git add go.mod go.sum
            fi
            popd
```

.github/workflows/testing.yaml
```diff
@@ -12,7 +12,65 @@ concurrency:
...
+      - name: Run go work sync
+        if: env.UPDATE != ''
+        run: go work sync
+
+      - name: Commit and push changes
+        if: env.UPDATE != ''
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+          git commit -m "fix: sync dependencies with go mod tidy" || echo "No changes to commit"
```
erdii, line 68:
```suggestion
          git commit -m "fix: sync dependencies with go mod tidy / go work sync" || echo "No changes to commit"
```

.github/workflows/testing.yaml
```diff
@@ -12,7 +12,65 @@ concurrency:
...
+  update-go-modules:
+    if: github.actor == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
erdii, line 24: Doesn't this break pushing the commit this job potentially creates?

https://git-scm.com/docs/shallow

> It is unclear how "push into/from a shallow repo" should behave.

erdii (changes requested):
I'd like to test this first before merging, but I'm not sure how to do.

Maybe a disconnected fork that doesn't have gh actions restrictions?

.github/workflows/testing.yaml
weherdh: What I know the `fetch-depth: 0` clones the full repo from the history, which can make the `git push` works normally. The GH actions default to a shallow clone  `fetch-depth: 1`, which may break the push, since the doc mentions: `It is unclear how "push into/from a shallow repo" should behave.`
https://github.com/actions/checkout
```
# Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

erdii: Ah - got it! :+1:

erdii (commented):
The changes look good!
I'm still unsure on how to test it though. :thinking:

weherdh:
This is the test from my forked repo:
1. The dependabot PR and last commit: https://github.com/weherdh/package-operator/pull/11#issuecomment-2670086722, and the action ran on this: https://github.com/weherdh/package-operator/actions/runs/13425145327/job/37506556120 (The lint-unit-int failed because my forked repo did not sync all the tags in this origin repo)
2. The normal PR: https://github.com/weherdh/package-operator/pull/116 Which will skip the update-go-modules action job

erdii (approved):

PR #1788, opened 2025-02-19
[PKO-226] Clean Up: Remove .status.phase from all APIs

Ankit152:
### Summary

Jira: [PKO-226](https://issues.redhat.com/browse/PKO-226)

### Change Type

* Cleanup

### Check List Before Merging

      standard.


thetechnick (changes requested):
There are some default directives that can also be removed now:
```
// +kubebuilder:default={phase:Pending}
```

Also I am curious, how do we replace the Phase-Based status for printing it via kubectl?
I'd like to make sure that we don't release a version of PKO that does not have any human readable status for it's CRs.

internal/cmd/client.go
```diff
@@ -35,7 +35,7 @@ func (c *Client) GetObjectset(ctx context.Context, name string, ns string) (*cor
 		return nil, fmt.Errorf("getting package objectsetlist : %w", err)
 	}
 	for i := range objreslist.Items {
-		if objreslist.Items[i].Status.Phase == "Available" && strings.Contains(
```
Ankit152, line 38: I think the removal of this logic is causing the issue. 
@erdii any thoughts?

internal/cmd/client.go
```diff
@@ -66,7 +66,7 @@ func (c *Client) GetClusterObjectset(ctx context.Context, name string,
 		return nil, fmt.Errorf("getting package objectsetlist : %w", err)
 	}
 	for i := range objreslist.Items {
-		if objreslist.Items[i].Status.Phase == "Available" && strings.Contains(objreslist.Items[i].Name, name) {
```
Ankit152, line 69: same here.

internal/cmd/client.go
erdii: I'd approach finding the answer to this by asking these questions - one by one:
1. What does the surrounding method do and what is its purpose?
2. What does this if condition do and what is its purpose?
3. What did `.Status.Phase` represent within the context of 1. and 2.?
4. Are there other ways to represent the answer of 3.?

Ankit152: Thanks for the set of question you gave for me to follow-up that helped me to arrive at an understanding.

Whenever the `.Status.Phase` is `Available` and other conditions are met, the function is returning the respective ObjectSet/ClusterObjectSet and if not, it will return an error.
The `.Status.Phase` is set to `Available` when all the `.Status.Conditions` are met with `Condition.Type` to `Available`.
Now that we are getting rid of `.Status.Phase`, how about we write a helper function which iterates through `.Status.Conditions` and checks if all of them are `Avavilable`? If yes, it will return true and mimic similar behavior.

```
func isAllConditionsAvailable(conditions []metav1.Conditions) bool {
    for condition := conditions {
        if condition.Type != "Available" {
            return false
        }
    }
    return true
}
```

This above helper function will be replaced as follows:
```
if isAllConditionsAvailable(objreslist.Items[i].Status.Conditions) && strings.Contains(objreslist.Items[i].Name, name) {
    obj = &corev1alpha1.ClusterPackage{
                    ObjectMeta: metav1.ObjectMeta{
		        Name: objreslist.Items[i].Name,
		    },
         }
}
```

erdii (approved):

PR #1798, opened 2025-02-25
[PKO-208] Add GH Action that runs go mod tidy on dependabot PRs

weherdh:
### Summary

This is to resolve https://issues.redhat.com/browse/PKO-208

### Change Type

Bug Fix

### Check List Before Merging

      standard.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -7,55 +7,128 @@ on:
   push:
     branches: [main]
 
+env:
+  GO_VERSION: '>=1.23.x'
+
 concurrency:
   group: '${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}'
   cancel-in-progress: true
 
 jobs:
-  lint-unit-int:
+  update-go-modules:
+    if: github.actor == 'dependabot[bot]'
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yaml
```diff
@@ -12,50 +12,120 @@ concurrency:
...
+      - name: Ensure we have the latest commits
+        if: github.actor == 'dependabot[bot]' && env.UPDATE != ''
+        run: |
+          git fetch origin "${{ github.head_ref }}"
+          git reset --hard "origin/${{ github.head_ref }}"
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '>=1.22'
```
eqrx, line 96:
```suggestion
          go-version: '>=1.23.x'
```

eqrx (changes requested):

.github/workflows/testing.yaml
weherdh: If we update the version here, what about the above one in the `update-go-modules`, should we update it as well?

erdii: Yes. Both should be kept in sync.

(Is it possible to use YAML Anchors with GH Actions, so we only need to specify the version once?)

PR #1809, opened 2025-03-03
[PKO-235] Remove unused webhook component

pbabic-redhat:
### Summary

Remove the unused `package-operator-webhook` component.

Webhook was introduced to enforce objectset immutability but it was never actually deployed in downstream. Now we use CEL and other constraints in the apiserver instead.

### Jira
[PKO-235](https://issues.redhat.com/browse/PKO-235)

### Change Type
Clean-up

### Check List Before Merging

      standard.

### Additional Information
Docs update PR: [#19](https://github.com/package-operator/website/pull/19)


erdii (changes requested):
Generally this looks good to me. I have a couple of suggestions.

I think the webhook servers in our controller managers can also be removed:
- https://github.com/pbabic-redhat/package-operator/blob/df61ec97b5c1165420ff1697dcb7d69757137e27/cmd/remote-phase-manager/main.go#L102
- https://github.com/pbabic-redhat/package-operator/blob/df61ec97b5c1165420ff1697dcb7d69757137e27/cmd/package-operator-manager/components/components.go#L102

erdii (changes requested):
What? I changed the review verdict from "comment" to "request changes" and VSCode threw away all of the comments I had made... :crying_cat_face: 


Anyhow, here I go again, just a little sloppier - take my tone with a grain of salt and as always, question how I use the word "should":
- Overall this looks very good. Thanks!
- I think the two webhook servers in the controller managers can also be removed.
- I think the immutability tests can be deduplicated twice:
  - hypershift and non-hypershift should run the same code. There is prior art in the other objectset tests for this.
  - ClusterObjectSet and ObjectSet tests should use the adapter from the objectset controller package.
- the waiter that waits for the hypershift aware pko to become available again before running the tests is a good catch! I think it should live in a dedicated preamble separated from other tests though.

PR #1887, opened 2025-03-20
fix error handling in unpackReconciler and packageDeployer

erdii:

### Change Type

Bug Fix

### Check List Before Merging

      standard.

### DIFF
--- a/internal/packages/internal/packagedeploy/deployer.go
+++ b/internal/packages/internal/packagedeploy/deployer.go
@@ -129,13 +129,15 @@ func (l *PackageDeployer) Deploy(
 	pkg, err := l.structuralLoader.LoadComponent(ctx, rawPkg, apiPkg.GetComponent())
 	if err != nil {
 		setInvalidConditionBasedOnLoadError(apiPkg, err)
+		// Explicitly do not return an error here to avoid re-pulling
+		// and deploying a broken package over and over again.
 		return nil
 	}
 
 	// Check constraints
 	if err := validateConstraints(ctx, l.uncachedClient, apiPkg, pkg.Manifest, env); err != nil {
 		setInvalidConditionBasedOnLoadError(apiPkg, err)
-		return nil
+		return err
 	}
 
 	// prepare package render/template context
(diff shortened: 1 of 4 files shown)


internal/packages/internal/packagedeploy/deployer.go
```diff
@@ -129,13 +129,13 @@ func (l *PackageDeployer) Deploy(
 	pkg, err := l.structuralLoader.LoadComponent(ctx, rawPkg, apiPkg.GetComponent())
 	if err != nil {
 		setInvalidConditionBasedOnLoadError(apiPkg, err)
-		return nil
+		return err
```
thetechnick, line 132: We have to be careful about this one.
Now a LoadError is requeued very fast until going into exponential backoff.
I think the .Reconcile function should still not return an error for load errors and instead use the backoff manager we are already using to slow down retrying image pulls.

erdii: Good catch!

One question though: IIUC and I revert to what you said then there is no backoff involved at all. The package gets pulled, structural load fails, the caller (unpack_reconciler) updates .status.unpackedHash and it will never be retried again because the assumption is, that the package will stay broken. Is that correct?

thetechnick: Yes.

In the beginning the idea was that if the Package is Invalid, we don't retry.
Reason being that users will have to provide a new package image/config anyway.

If we want to change that we could do something like this and add this to unpack_reconciler.go:
```
	if meta.IsStatusConditionTrue(*pkg.GetConditions(), corev1alpha1.PackageInvalid) {
		backoffID := string(pkg.ClientObject().GetUID())
		r.backoff.Next(backoffID, r.backoff.Clock.Now())
		backoff := r.backoff.Get(backoffID)
		return ctrl.Result{
			RequeueAfter: backoff,
		}, nil
	}
```
But I think we should only do this in cases where retrying has a benefit or reason, like e.g. when constraints on the cluster version failed.

erdii: Ack! Thanks for the additional context.

I think this is a discussion we can have separately from this PR, so I'll just revert this line and we should be good to go.

eqrx (approved):

PR #1888, opened 2025-03-20
Update dependencies

eqrx:
Update of dependencies and tools.

Problems with conversion-gen:
it removed the "--input-dirs" flag and replaced it by positional arguments.
it has remove the "-h" flag.
it has renamed the "--out-file-base" to "--output-file"
ithas remove the "--output-base" flag


eqrx:
Superseeded

PR #1892, opened 2025-03-20
Ownerhandling

eqrx:


internal/packages/internal/packagedeploy/deployment_reconciler.go
```diff
@@ -291,10 +292,10 @@ func (r *DeploymentReconciler) reconcileSliceWithCollisionCount(
 
 	// Try to create slice
 	err := r.client.Create(ctx, slice.ClientObject())
-	if err == nil {
+	switch {
+	case err == nil:
 		return nil
-	}
-	if err != nil && !apimachineryerrors.IsAlreadyExists(err) {
+	case !apimachineryerrors.IsAlreadyExists(err):
```
erdii, line 298: I don't think adding a switch makes this more readable.
To the contrary - before it was "create, exit early if there was no error, otherwise: continue with this other stuff".

I propose replacing:
```golang
if err != nil && !apimachineryerrors.IsAlreadyExists(err) {
```
with
```golang
// Add newline here
if !apimachineryerrors.IsAlreadyExists(err) {
```
instead and leaving the rest as is.

erdii (changes requested):
overall lgtm

internal/packages/internal/packagedeploy/deployment_reconciler.go
erdii: And/or adding a comment that explains the behavior.

erdii (approved):

PR #1904, opened 2025-03-21
Lint

eqrx:
Needs #1902 first


erdii (approved):

eqrx:
This pr needs go1.24. After this pr was created we pinned to go1.23 -> pr is not mergable for now

PR #1951, opened 2025-04-07
fix: remove update-go-modules job from testing workflow because we couldn't make it work >:-|

erdii:


erdii:
fixed in https://github.com/package-operator/package-operator/pull/1953

PR #1952, opened 2025-04-08
[PKO-255]Fix to check out use PAT token

weherdh:
### Summary
This is a minor fix to use PAT token to check out

### Change Type

Bug Fix

### Check List Before Merging

      standard.

### DIFF
--- a/.github/workflows/testing.yaml
+++ b/.github/workflows/testing.yaml
@@ -7,6 +7,9 @@ on:
   push:
     branches: [main]
 
+env:
+  GO_VERSION: '>=1.24.x'
+
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yaml
```diff
@@ -7,6 +7,9 @@ on:
   push:
     branches: [main]
 
+env:
```
eqrx, line 10: What does the move of the go version into an env var fix?

eqrx (changes requested):

erdii:
I've taken the liberty to take the actual fix out of this PR and apply it directly in https://github.com/package-operator/package-operator/pull/1953 .

Thanks @weherdh the approach seems to work!

PR #1957, opened 2025-04-09
[PKO-245] Sanitize code

trevorwilliams2025:
### Summary

Currently the job runs go mod tidy in the submodules.

If git diff reports that the repo is dirty, the pipeline fails. Currently the git diff output is called with the "–quiet" flag which makes debugging hard. Remove flag and tidy up code.
### Change Type

Bug Fix

### Check List Before Merging

      standard.

### DIFF
--- a/cmd/build/ci.go
+++ b/cmd/build/ci.go
@@ -27,10 +27,15 @@ func (ci *CI) Lint(_ context.Context, _ []string) error {
 // PostPush runs autofixes in CI and validates that the repo is clean afterwards.
 func (ci *CI) PostPush(ctx context.Context, args []string) error {
 	self := run.Meth1(ci, ci.PostPush, args)
-	if err := mgr.ParallelDeps(ctx, self,
+	if err := mgr.SerialDeps(ctx, self,
 		run.Meth(generate, generate.All),
 		run.Meth(lint, lint.glciFix),
 		run.Meth(lint, lint.goModTidyAll),
+		run.Meth(lint, lint.goWorkSync),
+	); err != nil {
+		return err
+	}
+	if err := mgr.ParallelDeps(ctx, self,
 		run.Meth(lint, lint.govulnCheck),
 	); err != nil {
 		return err
--- a/cmd/build/ci.go
+++ b/cmd/build/ci.go
@@ -30,8 +30,8 @@ func (ci *CI) PostPush(ctx context.Context, args []string) error {
 	if err := mgr.SerialDeps(ctx, self,
 		run.Meth(generate, generate.All),
 		run.Meth(lint, lint.glciFix),
-		run.Meth(lint, lint.goModTidyAll),
 		run.Meth(lint, lint.goWorkSync),
+		run.Meth(lint, lint.goModTidyAll),
 	); err != nil {
 		return err
 	}
--- a/cmd/build/dev.go
+++ b/cmd/build/dev.go
@@ -29,9 +29,10 @@ func (dev *Dev) PreCommit(ctx context.Context, args []string) error {
 		return err
 	}
 
-	return mgr.ParallelDeps(ctx, self,
+	return mgr.SerialDeps(ctx, self,
 		run.Meth(lint, lint.glciFix),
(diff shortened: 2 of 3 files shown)


cmd/build/ci.go
```diff
@@ -27,10 +27,15 @@ func (ci *CI) Lint(_ context.Context, _ []string) error {
 // PostPush runs autofixes in CI and validates that the repo is clean afterwards.
 func (ci *CI) PostPush(ctx context.Context, args []string) error {
 	self := run.Meth1(ci, ci.PostPush, args)
-	if err := mgr.ParallelDeps(ctx, self,
+	if err := mgr.SerialDeps(ctx, self,
 		run.Meth(generate, generate.All),
 		run.Meth(lint, lint.glciFix),
 		run.Meth(lint, lint.goModTidyAll),
+		run.Meth(lint, lint.goWorkSync),
```
erdii, line 34: I'm not sure about the ordering here. @weherdh built it exactly the other way round [1] and first calls `go work sync`, then `go mod tidy` in all modules.

Which ordering is correct? :thinking: 

[1] https://github.com/package-operator/package-operator/blob/ad5a76427ce069a9ac4bc88dab02c61ea11df534/.github/workflows/testing.yaml#L34-L44

cmd/build/dev.go
```diff
@@ -29,9 +29,10 @@ func (dev *Dev) PreCommit(ctx context.Context, args []string) error {
 		return err
 	}
 
-	return mgr.ParallelDeps(ctx, self,
+	return mgr.SerialDeps(ctx, self,
 		run.Meth(lint, lint.glciFix),
 		run.Meth(lint, lint.goModTidyAll),
+		run.Meth(lint, lint.goWorkSync),
```
erdii, line 35: Same here.

cmd/build/ci.go
```diff
@@ -27,10 +27,15 @@ func (ci *CI) Lint(_ context.Context, _ []string) error {
 // PostPush runs autofixes in CI and validates that the repo is clean afterwards.
 func (ci *CI) PostPush(ctx context.Context, args []string) error {
 	self := run.Meth1(ci, ci.PostPush, args)
-	if err := mgr.ParallelDeps(ctx, self,
+	if err := mgr.SerialDeps(ctx, self,
```
erdii, line 30: The need for serial calls is a good catch!

cmd/build/ci.go
Ajpantuso: I asked @trevorwilliams2025 to conduct his own tests, but my quick assessment is that `go work sync` should run after `go mod tidy`.

If it runs before essentially all it is doing is ensuring that the `replace` directives are persisted for modules depending on other local modules in the workspace.

If it runs _after_ `go mod tidy` on each module it will also update transitive dependencies for modules depending on other local modules in the workspace.

So:
IF Module A depends on Module B
AND Module B updates direct dependency C
AND `go mod tidy` updates transitive dependency D in Module B
THEN `go work sync` will update both dependency C and D in Module A

Ajpantuso: Actually `go work sync` doesn't even ensure the replace directives.
It probably would ensure `replace` directives, but again in that case having the modules tidied first is probably better.

weherdh: I tested that in the dependabot, and my testing result is `go work sync` should run before the `go mod tidy`.
Otherwise the CI will not be clean since `../do CI:PostPush` will run go mod tidy clean the modules and there are always dependencies that need to be clean

trevorwilliams2025: matching the work @weherdh has completed for code consistency, i was unable to produce issues either way during testing.

cmd/build/ci.go
trevorwilliams2025: Might help explain inconsistencies

cmd/build/dev.go
trevorwilliams2025: making order consistent with existing code.

erdii (approved):

PR #1965, opened 2025-04-16
[PKO-243]feat:Expose log level configuration to package configuration

weherdh:
### Summary

This is to resolve PKO-243

### Change Type

 New Feature 

### Check List Before Merging

      standard.


config/packages/package-operator/components/hosted-cluster/manifest.yaml.tpl
```diff
@@ -27,6 +27,9 @@ spec:
   config:
     openAPIV3Schema:
       properties:
+        logLevel:
+          description: Log levels for package operator
```
pbabic-redhat, line 31: It think it might be better to use the same description as in the CLI flags for consistency. But feel free to disagree. :slightly_smiling_face: 
```suggestion
          description: Log level. Default is -1 (warn). Higher numbers increase verbosity (e.g., 0 = info, 1 = debug)
```

pbabic-redhat (changes requested):
Hey @weherdh 
I submitted a partial review (didn't have time to go through all of it today) so that we can iterate more efficiently. I left a couple of suggestions (minor changes). :slightly_smiling_face:

weherdh:
> Hey @weherdh I submitted a partial review (didn't have time to go through all of it today) so that we can iterate more efficiently. I left a couple of suggestions (minor changes). 🙂

Thanks a lot, that's all valid inputs, I have addressed them all

internal/constants/const.go
```diff
@@ -15,4 +15,7 @@ const (
 	FieldOwner = "package-operator"
 	// OwnerStrategyAnnotationKey is the k8s annotation key that denotes the owner of a resource.
 	OwnerStrategyAnnotationKey = "package-operator.run/owners"
+
+	// Verbosity level for Info logs that are hidden by default.
+	LogLevelInfo = 1
```
pbabic-redhat, line 20: This name is misleading, logs with `log.V(1)` don't actually show up on info level (`LOG_LEVEL=0`), they show up one level higher, so this should be called something like `LogLevelDebug`.

pbabic-redhat (changes requested):
Sorry for the delay, I left a single comment. I'm going to do some final testing, otherwise looks good. :slightly_smiling_face:

pbabic-redhat (approved):
Looks good, lgtm! :rocket:

PR #1997, opened 2025-05-20
[PKO-265]Expose HostedCluster.Spec.NodeSelector to package templates

weherdh:
### Summary

This is to resolve PKO-265

### Change Type

New Feature

### Check List Before Merging

      standard.


integration/package-operator/hypershift_test.go
```diff
@@ -106,26 +107,96 @@ func TestHyperShift(t *testing.T) {
...
+				client.ObjectKey{Name: pod.Spec.NodeName},
+				node))
+
+			// Verify node has matching labels
+			for k, v := range testNodeSelector {
+				assert.Equal(t, v, node.Labels[k],
+					"Pod %s/%s scheduled on node %s missing required label %s=%s",
+					pod.Namespace, pod.Name, node.Name, k, v)
+			}
+		}
```
erdii, line 198: Verifying that the deployment has the correct selector in .spec.template.spec.nodeSelector should be enough when coupled with a `Waiter.WaitForCondition(ctx, hcDeployment, appsv1.DeploymentAvailable, metav1.ConditionTrue, wait.WithTimeout(300*time.Second))`.

```suggestion
```

erdii (changes requested):
Small nit to reduce amount of test code + api docs need to be regenerated.

erdii (approved):

PR #2025, opened 2025-06-10
[PKO-280]Prevent discarding of log messages due to missing logger in ctx

weherdh:
### Summary

This is to resolve PKO-280

### Change Type

Bug Fix 

### Check List Before Merging

      standard.

### DIFF
--- a/cmd/package-operator-manager/bootstrap/bootstrap_test.go
+++ b/cmd/package-operator-manager/bootstrap/bootstrap_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/go-logr/logr"
+
 	"github.com/go-logr/logr/testr"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
@@ -94,7 +96,11 @@ func TestBootstrapper_bootstrap(t *testing.T) {
 	t.Parallel()
 
 	c := testutil.NewClient()
-	b := &Bootstrapper{client: c}
+	// Create a test logger
+	logger := testr.New(t)
+	b := &Bootstrapper{
+		client: c,
+	}
 
 	var (
 		runManagerCalled bool
@@ -124,8 +130,9 @@ func TestBootstrapper_bootstrap(t *testing.T) {
 		}).
 		Return(nil)
 
-	ctx, cancel := context.WithTimeout(
-		context.Background(), testBootstrapTimeout)
+	// Create a context with the logger
+	ctx := logr.NewContext(context.Background(), logger)
+	ctx, cancel := context.WithTimeout(ctx, testBootstrapTimeout)
 	defer cancel()
 	err := b.bootstrap(ctx, func(ctx context.Context) error {
 		runManagerCalled = true
--- a/cmd/package-operator-manager/main.go
+++ b/cmd/package-operator-manager/main.go
@@ -49,14 +49,16 @@ func run(opts components.Options) error {
 		Development: false,
 		Level:       zapLevel,
 	}
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&zapOpts)))
+	logger := zap.New(zap.UseFlagOptions(&zapOpts))
+	ctrl.SetLogger(logger)
 	if opts.PrintVersion != nil {
 		_ = version.Get().Write(opts.PrintVersion)
 
 		return nil
 	}
 
-	ctx := logr.NewContext(ctrl.SetupSignalHandler(), ctrl.Log)
+	// Use the explicitly created logger instance to ensure we have a valid logger in context
+	ctx := logr.NewContext(ctrl.SetupSignalHandler(), logger)
 	if len(opts.SelfBootstrap) > 0 {
 		if err := di.Provide(bootstrap.NewBootstrapper); err != nil {
 			return err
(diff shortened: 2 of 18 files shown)


internal/preflight/dryrun.go
```diff
@@ -19,9 +19,15 @@ import (
 
 type DryRun struct {
 	client client.Writer
+	log    logr.Logger
 }
 
-func NewDryRun(client client.Writer) *DryRun { return &DryRun{client: client} }
+func NewDryRun(client client.Writer, log logr.Logger) *DryRun {
```
eqrx, line 25: I see that you added a Logger as parameter to several functions that had no Context. While I prefer that instead of just hiding it in the Context I would like to discuss how we can make this consistent. I would argue that it is better to either:
- Only get the Logger from Context. If a method needs a logger but it gets no Context it needs to be rewritten to receive a context.
- Always pass the logger explicitly, never getting it from the Context.

What do you think? (also @erdii )

eqrx (changes requested):

internal/preflight/dryrun.go
weherdh: If we use option 1 here, what is the exact update we can have? Is this the correct way to update it?
```
log, err := logr.FromContext(ctx)
	if err != nil {
		return fmt.Errorf("no logger in context")
	}
```
For the controller runtime, we can use `ctrl.LoggerFrom(ctx)`, but here if we want to solve the missing logger to inject the log is a better way?

erdii (commented):
I've added some comments on some patterns - didn't repeat the comments for the full PR.

cmd/package-operator-manager/main.go
```diff
@@ -49,14 +49,16 @@ func run(opts components.Options) error {
...
+	ctrl.SetLogger(logger)
 	if opts.PrintVersion != nil {
 		_ = version.Get().Write(opts.PrintVersion)
 
 		return nil
 	}
 
-	ctx := logr.NewContext(ctrl.SetupSignalHandler(), ctrl.Log)
+	// Use the explicitly created logger instance to ensure we have a valid logger in context
+	ctx := logr.NewContext(ctrl.SetupSignalHandler(), logger)
```
erdii, line 61: What do we get from this change?

cmd/package-operator-manager/main.go
```diff
@@ -77,7 +79,17 @@ func run(opts components.Options) error {
...
-		if err := bs.Bootstrap(ctx, func(ctx context.Context) error {
+		if err := bs.Bootstrap(ctx, func(bootstrapCtx context.Context) error {
+			_, err := logr.FromContext(bootstrapCtx)
+			if err != nil {
+				// Only get logger from original context and add it to bootstrapCtx if bootstrapCtx doesn't have one
+				logger, err := logr.FromContext(ctx)
+				if err != nil {
+					return fmt.Errorf("getting logger from context in bootstrap function: %w", err)
+				}
+				bootstrapCtx = logr.NewContext(bootstrapCtx, logger)
```
erdii, line 90: This procedure is not necessary because bs.Bootstrap calls the callback with a child of the context that is passed in.

I'd argue that this should just return `err` when it's not nil.

internal/controllers/objectsets/remotephase_reconciler.go
```diff
@@ -51,7 +52,7 @@ func (r *objectSetRemotePhaseReconciler) Teardown(
 	ctx context.Context, objectSet adapters.ObjectSetAccessor,
 	phase corev1alpha1.ObjectSetTemplatePhase,
 ) (cleanupDone bool, err error) {
-	log := logr.FromContextOrDiscard(ctx)
+	log := ctrl.LoggerFrom(ctx)
```
erdii, line 55: ctrl.LoggerFrom does a silent fallback to a default logger in https://github.com/kubernetes-sigs/controller-runtime/blob/4c2e9cec2954e6d5a9522c65b2e22f24acb750b5/pkg/log/log.go#L87

Is this what we want here?

cmd/package-operator-manager/bootstrap/bootstrap_test.go
```diff
@@ -94,7 +96,12 @@ func TestBootstrapper_bootstrap(t *testing.T) {
 	t.Parallel()
 
 	c := testutil.NewClient()
-	b := &Bootstrapper{client: c}
+	// Create a test logger and add it to the bootstrapper
+	logger := testr.New(t)
+	b := &Bootstrapper{
+		client: c,
+		log:    logger,
```
erdii, line 103: What do you think about dropping the field Bootstrapper.log completely because the context that is passed to Bootstrapper.bootstrap already contains a logger?

cmd/package-operator-manager/main.go
weherdh: So from my understanding, the above `ctrl.SetLogger(logger)` already set `ctrl.Log = logger`. Here the change is to make the intention clear and maintain consistency.

internal/controllers/objectsets/remotephase_reconciler.go
weherdh: From my understanding:
1. Using `ctrl.LoggerFrom(ctx)` aligns with our requirement to avoid discarding logs. It ensures that even if a logger is missing from the context, the logs will still be directed to the global controller-runtime logger.
2. In my opinion, this is the safest way to avoid discarding logs. It is very hard to trace back every `ctx` whether there is a logger injected, even our unit test and integration test are all passed, it still cannot guarantee every `ctx` we used has a logger, and it can be very risky and lead to breaking change if we only use `logr.FromContext(ctx)` and return panic

eqrx:
Closing since original issue was marked as won't do

PR #1927, opened 2025-04-02
fix: garbage collect metrics for deleted objects

robshelly:
### Summary
Fix bug in which metrics continue to be exported for packages and objectsets which have been deleted. 

### Change Type

Bug Fix

### Check List Before Merging

- [x] This PR passes all pre-commit hook validations.
- [x] This PR is fully tested and regression tests are included.
- [x] Relevant documentation has been updated.
- [x] The commits in this PR follow the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)
      standard.

PR #2028, opened 2025-06-17
fix(ObjectTemplate controller): copy finalizers from existing templated object to prevent removing them

erdii:
this was causing issues with Packages owned by ObjectTemplates because there was a hot-loop to add and remove the metrics finalizer that was recently added in https://github.com/package-operator/package-operator/pull/1927

Co-Authored-By: Petr Babic <pbabic@redhat.com>

### Check List Before Merging

      standard.


robshelly (approved):
LGTM!

PR #2477, opened 2026-03-16
[PKO-341] Use RevisionEngine in ObjectSetReconciler

pbabic-redhat:
### Summary

Replaces the PhaseReconciler with Boxcutter's reconcilers in the ObjectSetController.

Jira: [PKO-341](https://redhat.atlassian.net/browse/PKO-341)

### Check List Before Merging

      standard.


pbabic-redhat:
@coderabbitai review

pbabic-redhat:
@coderabbitai review

internal/controllers/controllers.go
```diff
@@ -191,3 +197,122 @@ func RemoveDynamicCacheLabel(
...
+			return err
+		}
+		var objectConditions []metav1.Condition
+		if err := json.Unmarshal(j, &objectConditions); err != nil {
+			return err
+		}
+
+		var conditionMappings []corev1alpha1.ConditionMapping
+		for _, objectsetobject := range ownerObjects {
+			if objectsetobject.Object.GetName() == obj.GetName() {
```
erdii, line 284: **Bug: Condition mapping matches objects by name only, ignoring GVK** (CONFIRMED)

This loop matches `ownerObjects` to actual objects by `GetName()` only, without checking GroupVersionKind or namespace. It also doesn't `break` on the first match, so the last name-colliding entry's `ConditionMappings` silently win.

**Failure scenario:** An ObjectSet with a ConfigMap named `config` and a Secret named `config` across phases — the Secret's ConditionMappings overwrite the ConfigMap's.

**Regression:** The old `mapConditions` function associated condition mappings per-object directly (no name-based lookup), so there was no ambiguity.

Consider matching on GVK + namespace + name, and adding a `break` after the first match.

erdii:
these have yet to be reviewed by me:

## Code Review Findings (max effort)

> Note: This PR was developed against an unreleased version of boxcutter checked out at `../boxcutter-review`.

### CONFIRMED — Correctness Bugs

**1. Inverted boolean return in namespace-deletion teardown** — `remotephase_reconciler.go:107`

`return err != nil, err` returns `cleanupDone=false` on Update success and `cleanupDone=true` on failure — the opposite of correct. Should be `return err == nil, err`. Tests miss this because they discard the `cleanupDone` return value with `_`.

---

**2. Condition mapping matches objects by name only, ignoring GVK** — `controllers.go:284`

`MapConditionsToObjectSetOrPhase` matches `ownerObjects` to actual objects by `GetName()` only. No GVK or namespace check, and no `break` on match — last same-named entry wins. Two objects with the same name but different kinds (e.g., ConfigMap `config` and Secret `config`) cause the first one's ConditionMappings to be silently dropped. **Regression:** the old per-object `mapConditions` had no such ambiguity.

---

**3. CSA field manager migration narrowed from 3 to 1** — `phase_engine_factory.go:89`

The old `phaseReconciler` migrated CSA managed fields from `{'package-operator', 'package-operator-manager', 'remote-phase-manger'}`. Boxcutter only migrates from the single `FieldOwner` (`package-operator`). Objects managed by the two legacy field owners will hit SSA field ownership conflicts on upgrade.

---

**4. CreateCollisionError swallowed — no status condition** — `objectsetphase_reconciler.go:132`

`CreateCollisionError` is caught and returned as `(Result, nil)`. Since err is nil, the controller skips `UpdateObjectSetOrPhaseStatusFromError` (which sets `CollisionDetected`). Users get silent backoff retries with no status visibility. The ObjectSet path does NOT catch this error, so collisions surface correctly there — asymmetry.

---

**5. ForceAdoption override missing in ObjectSetPhase path** — `objectsetphase_reconciler.go:111`

`PKO_FORCE_ADOPTION` env var and `package-operator` package label overrides are only applied in the ObjectSet path (`objectsetphases_reconciler.go:290-292`), not the ObjectSetPhase controller path. **Regression:** the old `defaultAdoptionChecker.Check()` applied these overrides for all paths.

---

**6. AddDynamicCacheLabel changed from MergeFrom to Merge** — `controllers.go:177`

Changed from `client.MergeFrom(obj)` (diff-only patch) to `client.Merge` (sends full object as patch body). Can silently overwrite concurrent modifications. `RemoveDynamicCacheLabel` at line 194 still correctly uses `client.MergeFrom`. **Regression** from old code.

---

### PLAUSIBLE — Likely Issues

**7. GetValidationError reports probe failures as PhaseValidationError** — `remotephase_reconciler.go:262`

`remotePhaseResult.GetValidationError()` wraps probe failures (e.g., "no status reported") as `PhaseValidationError`. This violates boxcutter's PhaseResult contract where `GetValidationError` represents preflight validation failures. Status messages will mislead users with "Validation Errors" for connectivity issues.

---

**8. Wrong validator for cluster-scoped controller** — `objectsetphase_controller.go:148`

`NewSameClusterClusterObjectSetPhaseController` uses `NewNamespacedPhaseValidator` instead of `NewClusterPhaseValidator`. All other cluster-scoped constructors use `NewClusterPhaseValidator`. Currently benign (short-circuited by empty-namespace early return) but latent bug.

---

**9. TranslateCollisionProtection panics on unknown values** — `translation.go:21`

`default: panic(...)` for unknown `CollisionProtection` values. The CRD has no `+kubebuilder:validation:Enum` annotation, so the API server accepts arbitrary strings. The old code treated unknown values as `Prevent` (safe default). **Regression** in robustness.

---

**10. updateStatus type-asserts `*remotePhaseResult`** — `objectsetphases_reconciler.go:184`

Fragile type assertion to extract `controllerOf`. If boxcutter ever wraps/decorates `PhaseResult` values, the assertion silently fails and remote phase controller references are dropped from status — permanently reporting ObjectSet as InTransition.

---

**11. SetNamespace/SetLabels mutates shared unstructured map** — `objectsetphases_reconciler.go:273`

`buildRevision`'s range loop copies `ObjectSetObject` structs, but the underlying `unstructured.Unstructured.Object` map is shared. `SetNamespace` and `SetLabels` contaminate the informer cache. Currently idempotent but a latent defect.

---

### Cleanup

**12. `os.Getenv` called in nested loop** — `objectsetphases_reconciler.go:290`

`os.Getenv(constants.ForceAdoptionEnvironmentVariable)` runs N×M times (phases × objects) per reconcile. It's a process-level constant — read once at init.

**13. Dead field: `remotePhase`** — `objectsetphases_reconciler.go:42`

Stored at construction but never read. All remote phase handling goes through `remoteEnabledPhaseEngine` inside the revision engine factory chain.

PR #2516, opened 2026-04-22
Introduce `./do dev:IntegrationDirect` to allow partially running integration tests without bootstrapping a full integration test environment

erdii:
## Introduce `./do dev:IntegrationDirect` to allow partially running integration tests without bootstrapping a full integration test environment

This can be useful when rapidly iterating on code changes and runnning package-operator-manager outside of the dev cluster with `./do dev:run`.

After this change devs can restart `./do dev:run` and then use `./do dev:integrationdirect` to only run a specific integration-test.

`dev:integrationdirect` will use go tests' short mode to skip the upgrade test, because it will install and upgrade pko inside the integration cluster which is undesirable when running pko-manager out-of-cluster.

Example: `./do dev:integrationdirect TestObjectSet_handover`

---

The PR also contains a change to the routine that detects package-operator's namespace before starting integration tests and now returns hardcoded values for kubernetes and openshift environments:
- kubernetes: `package-operator-system`
- openshift: `openshift-package-operator`

---

The PR also contains a change to all routines that detect if a certain api is installed in the apiserver to use `errors.Is(err, &discovery.ErrGroupDiscoveryFailed{})` instead of `discovery.IsGroupDiscoveryFailedError(errors.Unwrap(err)` to prevent not detecting errors that are wrapped into other errors more than one layer.
It does not do anything useful for now, but it may prevent nasty surprises in the future.

### Change Type

New Feature

### Check List Before Merging

      standard.

### DIFF
--- a/cmd/build/dev.go
+++ b/cmd/build/dev.go
@@ -84,6 +84,20 @@ func (dev *Dev) Integration(ctx context.Context, args []string) error {
 	return test.Integration(ctx, false, filter)
 }
 
+// IntegrationDirect runs local integration tests in a KinD cluster without setting up the cluster first.
+func (dev *Dev) IntegrationDirect(ctx context.Context, args []string) error {
+	var filter string
+	switch len(args) {
+	case 0:
+		// nothing
+	case 1:
+		filter = args[0]
+	default:
+		return errors.New("only supports a single argument")
+	}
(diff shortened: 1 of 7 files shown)


erdii:
@coderabbitai full review

pbabic-redhat (approved):
Yes please! 🚀

